PR #3, opened 2015-02-25
Add OSX support

Noltari:
Signed-off-by: Álvaro Fernández Rojas <noltari@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,16 @@ SBINDIR = $(PREFIX)/sbin
 DOCDIR = $(PREFIX)/share/doc
 MANDIR = $(PREFIX)/share/man
 
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Darwin)
+  CFLAGS += -D__osx__
+  LDLIBS += -liconv
+endif
+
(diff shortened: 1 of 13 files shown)


andreasbombe:
Thank you. I have already planned to remove the Linux dependencies. However, I plan to do my own data structure definitions rather than just copy in the Linux include files which were just used way back when by someone to save some typing work.

This means I will not pull this patch as-is, but I welcome you to adapt your fixes after I've gotten rid of the Linux include dependencies.

Noltari:
Ok, I'm looking forward to that changes and I will be glad to help you if I can.
I personally need dosfstools to be compatible with OS X in order to be able to build OpenWrt for the Raspberry PI on OS X.

Noltari:
BTW, what about something like this? https://github.com/Noltari/dosfstools/commit/f91225c9e48de1ff339193991ca5e1acacbc6f7f

Noltari:
Closing, see https://github.com/dosfstools/dosfstools/pull/4, https://github.com/dosfstools/dosfstools/pull/5 and https://github.com/dosfstools/dosfstools/pull/6

OS X support will be added on another PR once those PR are merged.

PR #5, opened 2015-02-26
Remove linux includes and int types

Noltari:


andreasbombe:
I see you've found I've merged your other pull requests. I found out I don't like the merge flow github offers through the web interface.

I also merged the non standard int types removal patch (slightly changed to keep comments aligned) from this series.

I'll also take the linux/msdos_fs.h removal with slight changes (keep attribute packed and name/ext split) – I could just add these myself into your patch, it's probably overkill to have you spin a new patch.

The rest is not quite useful. You just pull in the Linux specific include files, but these are used because they define Linux specific ioctls which would not exist on other platforms unless they run some sort of Linux compatibility layer. Sure, it compiles and since the information gained from the ioctl calls is not strictly necessary it will work just fine if the invalid ioctl calls error out.

This will need better handling, ideally using equivalents on the other platforms or at least encapsulate this stuff in Linux only parts. I will do something for that.

PR #7, opened 2015-03-07
Add OSX and FreeBSD support

Noltari:
Introduce changes requested in: https://github.com/dosfstools/dosfstools/pull/5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,11 +28,18 @@ DOCDIR = $(PREFIX)/share/doc
 MANDIR = $(PREFIX)/share/man
 
 #OPTFLAGS = -O2 -fomit-frame-pointer -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64
-OPTFLAGS = -O2 -fomit-frame-pointer -D_GNU_SOURCE $(shell getconf LFS_CFLAGS)
+OPTFLAGS = -O2 -fomit-frame-pointer -D_GNU_SOURCE
 #WARNFLAGS = -Wall -pedantic -std=c99
 WARNFLAGS = -Wall -Wextra -Wno-sign-compare -Wno-missing-field-initializers -Wmissing-prototypes -Wstrict-prototypes -Wwrite-strings
 DEBUGFLAGS = -g
 CFLAGS += $(OPTFLAGS) $(WARNFLAGS) $(DEBUGFLAGS)
+
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Darwin)
+  LDLIBS += -liconv
+else
+  OPTFLAGS += $(shell getconf LFS_CFLAGS)
+endif
 
 VPATH = src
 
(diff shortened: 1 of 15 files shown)


Noltari:
Updated PR with changes requested in https://github.com/dosfstools/dosfstools/pull/5
- msdos_fs.h (attribute packed and name/ext split)
- Add compatible ioctl calls for OSX and FreeBSD where posible
- FreeBSD support is untested (changes based on available documentation)

andreasbombe:
I merged the linux/msdos_fs.h removal, thanks.

However, when I reviewed the changed for the ioctl patch I found that the Linux dependencies go much deeper. Note that the ioctls are done on certain hardcoded major and minor numbers of block devices and I'm that doesn't work like it is supposed to outside of Linux. In fact, it has even been unreliable on Linux for many years since the st_rdev field is tested on an assumption of 8 bit major/minor device numbers.

I intend to revamp and move that detection out into its own file, but likely only after making a release with the current changes. Either I'll create a branch or only commit it after the release is done. Please don't waste your effort on the current state of things (it's not really wasted though, the changes should be portable to what I have in mind) — I'll drop you a note when I have something to show.

Noltari:
Ok, I'll be looking forward for your note.

Thanks again for your interest in getting support for non Linux OS ;).

andreasbombe:
It took much longer than anticipated but I have finally pushed all the changes to get rid of the device number problems and make it not depend on Linux specifics. It does require libblkid now but I think that is available at least on FreeBSD. The only really Linux dependent stuff is in `get_block_geometry()` in `device_info.c` and that is ifdef'd to a dummy on non-Linux systems. It's not really required for modern devices that would only present fake values anyway.

I'd be interested to hear if that is unsuitable and something else needs to be done.

Noltari:
Hello Andreas,

Unfortunately libblkid isn't avaiable on OS X.
I've been looking for a solution to this problem and I noticed that e2fsprogs uses a private blkid library for OS X systems, but it isn't exported...

Regards,
Álvaro.

andreasbombe:
Ugh, I overlooked that there are two versions of libblkid (util-linux and originally in e2fsprogs) and the e2fsprogs version doesn't have the interesting parts, and that's the version easily available on FreeBSD it appears.

So I guess I'll get rid of libblkid again and have more low level stuff in `device_info.c`. I think I can copy the important parts from [blkdev.c](https://github.com/karelzak/util-linux/blob/master/lib/blkdev.c) from util-linux which has some support for other operating systems.

andreasbombe:
Got rid of libblkid, now using a copy of the portable blkdev.c from util-linux. It calls only a few functions in device_info.c, but you might want to check if those can be improved. In particular, the get_geometry is a dummy on non-Linux platforms (but it's only relevant for old, small devices (including floppies).

andreasbombe:
It's working on FreeBSD but I can't test OS X. It's missing at least the mounted check for mkfs but there is a fallback of simply not checking.

So now I'm closing this again. If there is still something not working please open an issue since that's where this discussion now belongs. Unless you have patches to add the missing parts for OS X, then by all means open a new pull request. Thanks.

PR #14, opened 2015-11-07
Compile with musl libc

doughdemon:
musl (http://musl-libc.org) does not define the type loff_t in its headers, making compilation fail badly. We could use a configure test for loff_t. On the other hand, because we are using AC_SYS_LARGEFILE, loff_t should be the same as off_t anyway (assuming --disable-largefile is not used)


andreasbombe:
I already removed all `loff_t` instances in a commit which I hadn't pushed until now.

Thanks and sorry for the duplicate work.

Issue #15, opened 2015-11-13
typo


Behrooz-Amoozad:
"such filesystem will not propably mount"

andreasbombe:
I fixed this by rewording instead, since the original message wasn't written very well.

PR #20, opened 2016-01-08
Fix typo from issue #15

JacobCWard:
Corrected “propably” to “probably”


andreasbombe:
Thanks, but I decided to reword the message in question instead.

PR #23, opened 2016-03-10
Include fix, and add --without-udev option

jhol:


andreasbombe:
I have taken the configure option patch, though I had to reformulate the option help (the option is for without, so the help should also mention without). I have also rewritten the commit log (libudev was optional before). Thanks.

I see no reason to take the include path "fix" since it doesn't actually fix anything.

PR #27, opened 2016-05-06
src/device_info.c: Fix undefined PATH_MAX under musl

asergi:
In musl libc, PATH_MAX is defined in \<limits.h>.

### DIFF
--- a/src/device_info.c
+++ b/src/device_info.c
@@ -17,6 +17,7 @@
 */
 
 
+#include <limits.h>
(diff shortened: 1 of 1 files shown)


RomainNaour:
Hi,
Thanks for this patch!
I added it to dosfstools Buildroot package:
http://lists.busybox.net/pipermail/buildroot/2016-September/171354.html
Best regards,
Romain

andreasbombe:

PR #28, opened 2016-05-10
Don't link to udev via LDFLAGS. There's LDADD for such a task.

Polynomial-C:
See also:
https://www.gnu.org/software/automake/manual/html_node/Linking.html

Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


andreasbombe:

PR #31, opened 2016-06-30
Improve OS X endian compatibility and add 2 more files to .gitignore

Noltari:
Use autoconf to check for libkern/OSByteOrder.h existence instead of relying on __APPLE__ definition.
Add a couple of files to gitignore.


andreasbombe:

Noltari:
Thanks!

PR #36, opened 2016-10-12
Include limits.h for definition of PATH_MAX

f1rebird:
PATH_MAX is a macro defined in limits.h.
In order to work/compile with musl-libc, include the required header in src/device_info.c


f1rebird:
Never mind, seems to be already fixed.
Sorry for the noise.

PR #55, opened 2017-07-05
Fix reading and writing empty labels

pali:
Please properly review these changes as I'm not sure if there is no hidden threat. Such nonsense as padding with spaces, two different location for volume label, special first byte which indicate deleted entry or special string "NO NAME    " (instead of empty string) can be really found in FAT.

The good tool for reading label from root FAT directory is `mdir` from mtools. And for inseparability is good `mlabel` and recent `blkid`.

Seems we are missing tests for such FAT nonsenses...

### DIFF
--- a/src/boot.c
+++ b/src/boot.c
@@ -535,6 +535,8 @@ static void write_volume_label(DOS_FS * fs, char *label)
 	offset = alloc_rootdir_entry(fs, &de, label, 0);
     }
     memcpy(de.name, label, 11);
+    if (memcmp(de.name, "NO NAME    ", 11) == 0)
+        memcpy(de.name, "\xE5          ", 11);
     de.time = htole16((unsigned short)((mtime->tm_sec >> 1) +
 				       (mtime->tm_min << 5) +
 				       (mtime->tm_hour << 11)));
--- a/src/fatlabel.c
+++ b/src/fatlabel.c
@@ -81,12 +81,14 @@ int main(int argc, char *argv[])
 
     device = argv[1];
     if (argc == 3) {
-	strncpy(label, argv[2], 11);
 	if (strlen(argv[2]) > 11) {
 	    fprintf(stderr,
 		    "fatlabel: labels can be no longer than 11 characters\n");
 	    exit(1);
 	}
+	sprintf(label, "%-11.11s", argv[2]);
+	if (memcmp(label, "           ", 11) == 0 || label[0] == '\xE5')
+	    memcpy(label, "NO NAME    ", 11);
 	for (i = 0; label[i] && i < 11; i++)
 	    /* don't know if here should be more strict !uppercase(label[i]) */
 	    if (islower(label[i])) {
--- a/src/fatlabel.c
+++ b/src/fatlabel.c
@@ -105,10 +105,10 @@ int main(int argc, char *argv[])
 	read_fat(&fs);
     if (!rw) {
 	offset = find_volume_de(&fs, &de);
-	if (offset == 0)
-	    fprintf(stdout, "%.11s\n", fs.label);
-	else
+	if (offset != 0 && !IS_FREE(de.name))
 	    fprintf(stdout, "%.8s%.3s\n", de.name, de.name + 8);
+	else if (memcmp(fs.label, "NO NAME    ", 11) != 0)
+	    fprintf(stdout, "%.11s\n", fs.label);
 	exit(0);
     }
 
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -1481,6 +1481,8 @@ int main(int argc, char **argv)
 
 	case 'n':		/* n : Volume name */
 	    sprintf(volume_name, "%-11.11s", optarg);
+	    if (memcmp(volume_name, "           ", MSDOS_NAME) == 0 || volume_name[0] == '\xE5')
+	        memcpy(volume_name, NO_NAME, MSDOS_NAME);
 	    for (i = 0; volume_name[i] && i < 11; i++)
 		/* don't know if here should be more strict !uppercase(label[i]) */
 		if (islower(volume_name[i])) {


pali:
@andreasbombe: ping any comments for this pull request?

andreasbombe:
Sorry, I'm reviewing it only now.

A quick thing first, while I think I agree `mkfs` should accept the empty string for label name so as to avoid special casing when calling it from some script with the label name from a variable, I think `fatlabel` should error out in that case. I intend give `fatlabel` another option to clear the label (or reset the volume ID) rather than let it accept an empty string as a special case.

pali:
Ok, so I hope that first two commits are fine. Fourth commit is fine too?

And about setting empty label. Do not forget that label is always space padded to 11 characters. `` is same as ` ` and same as `           `. So if you really want to disabllow setting empty label via parameter, you need to correctly test all possible options. Also beware of the label argument which starts with 0xE5. As it is more compilcated I thought that it would be better and easier to specify empty label for erase label action.

andreasbombe:
I have taken the first, although this made me realize that more checking is needed for invalid characters beyond 0xe5 (and 0xe5 should be allowed and translated instead) but I will add that later.

The second one is where I will have separate function to remove the label instead – I don't want  to conflate the two, and it shouldn't have to allocate a directory entry only to mark it deleted.

I think I'll take the third (`sprintf` instead of `strncpy` is the right call) and later modify like the first.

The fourth points out a bug in that deleted directory entries were accepted as labels, but that should be fixed in `find_volume_de()`. And it changes the behavior of `fatlabel` in that it doesn't print anything when no label is found. So I'm not taking it.

But thanks for pointing out issues even with the patches I didn't take.

pali:
Hi @andreasbombe. Your fixes are not correct. In current master there are these problems:
1) fatlabel prints "NO NAME" when you erase label.
2) fatlabel does not handle empty/removed label from the root directory. Now function `find_volume_de` does not distinguish between facts "label in root directory is missing" and "label in root directory is empty".
3) fatlabel prefers label from the boot sector. But all other systems and tools prefers label from the root directory. Therefore if there is different label in boot sector and root directory, fatlabel outouts different string as all other tools. E.g. Windows XP when changing FAT label, it changes it **only** in root directory. And when you remove FAT label in Windows XP, then it is marked as removed only in root directory (in both cases boot sector is untouched).
4) fatlabel show label from boot sector padded to 11 characters with spaces.

pali:
And you forgot to update timestamp when calling `remove_label`.

andreasbombe:
@pali I can't follow your concerns:

1. `fatlabel` doesn't print anything when `--reset` is given.
2. `find_volume_de()` is supposed to find the label entry in the root directory. It used to have a bug where it returned an invalid entry in some cases instead of the correct or no entry. It doesn't look at the content, so if there's an empty label it will return it. I don't see what distinction is supposed to be lost.
3. `fatlabel` will only display the boot sector label if the root directory label is not found. IOW, it prefers the root directory label. And if Windows XP is indeed not updating the boot sector label Microsoft was violating their own published FAT32 specification.
4. It shows the whole label including spaces just as always.
5. What time stamp exactly?

pali:
Ok, I would try to explain it again with examples.

1.
```
$ ./src/fatlabel --reset /tmp/fat
$ ./src/fatlabel /tmp/fat
NO NAME    
$ blkid -p -s LABEL /tmp/fat

```
When FAT label is not set (= "NO NAME") then fatlabel show "NO NAME" instead of empty string or empty output. E.g. blkid handles it correctly. Also Windows XP show in this case empty label and not string "NO NAME".

2. Empty label is stored as: "NO NAME" in boot sector and "\xE5" in root directory. Please note that storing label into root directory is optional. Therefore there are 3 different states:

* label is missing in root directory
* "\xE5" in stored in root directory
* correct string is stored in root directory

Reading root directory should be handles as: if label is missing in root directory, then fallback to label from boot sector. If label in root directory is removed ("\xE5") then label is empty string independently of that is stored in boot sector. Otherwise filled string is label.

Test example:
```
./src/fatlabel /tmp/fat "NEW LABEL"
```
Then remove label on Windows XP and run fatlabel again. Currently it print:
```
$ ./src/fatlabel /tmp/fat
NEW LABEL
```
But it should not print "NEW LABEL" because it was changed -- removed in Windows XP. Windows XP just changed "NEW LABEL" in root directory to "\xE5" which means label is empty. Windows XP do not update label in boot sector, therefore after Windows XP change fat volume looked like:
boot sector = "NEW LABEL"
root dir = "\xE5"

3. "only display the boot sector label if the root directory label is not found" --> Not only if is not found, but also when label in root directory is empty (see 2.).

4. I do not think it is correct behavior to show those spaces. Windows XP, mtools and also blkid do not show them. Also linux kernel do not show file names padded with spaces. I consider this as a bug in fatlabel because it behave differently as all other existing software.

5. Timestamp for root dir label entry when --reset option is used.

pali:
And for example in 2., also mlabel show that label is empty.

andreasbombe:
I did my own experiments with Windows 10 and conclude that Windows completely ignores the existence of the boot sector label. When there's no root directory label it considers the filesystem unlabeled and when changing the label it doesn't touch the boot sector. Whereas when `fatlabel` doesn't find a root directory label it falls back to the boot sector label and so does blkid. Unfortunately this means that Linux and Windows disagree on the label in the specific case of no root directory label and boot sector label != "NO NAME".

1. It's what `fatlabel` has always done. I'm a bit reluctant to change how it works in case there's scripts depending on that (even though having "NO NAME" might not appear for unlabeled filesystems even then, since the boot sector label fields might be missing and then there's no output). I might add an option to explicitly switch to a different mode with "cleaner" output.

2. Empty label is stored as "NO NAME" in boot sector and not at all in root directory. There is no difference between deleted directory entries and non-existent directory entries. Expecting fields of deleted directory entries to have valid values is erroneous.

3. See 2.

4. See 1.

5. I don't see a particular value in setting a timestamp for a directory entry that is deleted anyway.

pali:
Non existance of the label in root directory is a very rare, more often label in root directory is marked as deleted (as 0xE5). So if there is inconsistency between different implementations, then it should be "optimized" for more common scenarios. It is known fact that Windows do not use label from boot sector (I think it is also written in some Microsoft documentation...)

It is possible to set label to "NO NAME" into root directory. And then it means that label is non-empty. Currently fatlabel is not able to distinguish between this and empty label state. mtools can.

Please note that GUI tools, like e.g. GParted do not use fatlabel, but rather mlabel just because of these bugs with handling empy label or different output (spaces, 0xE5, "NO NAME", ...). mtools/mlabel is more compatible with DOS/Windows machines for which FAT was is "home" filesystem (created by Microsoft).

On the other hand mtools is disaster for usage, it has uncommon arguments and non-linux way of usage. I (and probably other people too) would really approtiate if compatibility problems with other implementations in fatlabel are fixed. So dependency on mtools in more softwares could be dropped.

pali:
For 5. Reason is to know timestamp of the last change. When changing label to non-empty it is updated. But when setting label to empty, it is not updated... so it is some non-logic inconsistency.

PR #57, opened 2017-07-10
Fix signed integer overflow in FSTART

jwilk:
`uint16_t` was promoted to `int`, and then left shift could overflow it.
Add explicit cast to `uint32_t` to avoid undefined behavior.

Found using [american fuzzy lop](http://lcamtuf.coredump.cx/afl/).


andreasbombe:
Merged, thank you.

PR #61, opened 2017-08-31
src/check.c: Fix up mtools created bad dir entries

willnewton:
mtools writes uninitialized data to the case field of some
directory entries. Running fsck.fat on these filesystems
will cause the directory to get deleted which can lead to
data loss. Detect this situation and clear the flag instead.

mtools patch to fix the original issue:

https://lists.gnu.org/archive/html/info-mtools/2014-08/msg00000.html

Signed-off-by: Will Newton <willn@resin.io>

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -495,6 +495,13 @@ static int handle_dot(DOS_FS * fs, DOS_FILE * file, int dots)
 	    break;
 	}
     }
+    if (file->dir_ent.lcase & FAT_NO_83NAME) {
+	/* Some versions of mtools write these directory entries with random data in
+	   this field. */
+	printf("%s\n  Is a dot with no 8.3 name flag set, clearing.\n", path_name(file));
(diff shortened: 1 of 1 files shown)


dezgeg:
In addition to this, I see in my mtools-built images that the first cluster field of `.` and `..` are also incorrect:
````
/NIXOS/z088aj7gg4dw3rb3da1klf6jm7cyadxc-linux-4.12.8-dtbs/.
  Start (0) does not point to parent (10546)
/NIXOS/z088aj7gg4dw3rb3da1klf6jm7cyadxc-linux-4.12.8-dtbs/..
  Start (0) does not point to .. (15772)
````

floion:
@andreasbombe is this okay to merge from your side?

floion:
ping on this @andreasbombe

andreasbombe:
Ok, seems good, will merge this.

floion:
Thank you

PR #62, opened 2017-09-05
Fix integer overflow in check.c

newsoft:
The integer computation happening in date_dos2unix() conversion can overflow.

This should have no security impact, but any sane compiler/fuzzer will choke on this.

The suggested fix computes over 64-bit integers, then tries to downcast into a "time_t" if possible - defaulting to 0.

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -244,6 +244,7 @@ static time_t date_dos2unix(unsigned short time, unsigned short date)
 {
     int month, year;
     time_t secs;
+    long long int tempsecs;
 
     month = ((date >> 5) & 15) - 1;
     if (month < 0) {
@@ -251,11 +252,23 @@ static time_t date_dos2unix(unsigned short time, unsigned short date)
 	month = 0;
     }
     year = date >> 9;
-    secs =
+    tempsecs =
 	(time & 31) * 2 + 60 * ((time >> 5) & 63) + (time >> 11) * 3600 +
-	86400 * ((date & 31) - 1 + day_n[month] + (year / 4) + year * 365 -
+	(long long int)86400 * ((date & 31) - 1 + day_n[month] + (year / 4) + year * 365 -
 		 ((year & 3) == 0 && month < 2 ? 1 : 0) + 3653);
     /* days since 1.1.70 plus 80's leap day */
+
+    /* "time_t" is specified as "any arithmetic type" by POSIX, it can be as
+       small as a signed integer */
+    if (tempsecs > INT32_MAX) {
+      /* Date is completely invalid, return 01.01.1970 by default */
+      secs = (time_t)0;
+    }
+    else {
+      /* At this point it should be safe to cast into a "time_t" */
+      secs = (time_t)tempsecs;
+    }
+
     return secs;
 }
 


src/check.c
```diff
@@ -244,18 +244,31 @@ static time_t date_dos2unix(unsigned short time, unsigned short date)
...
+    tempsecs =
 	(time & 31) * 2 + 60 * ((time >> 5) & 63) + (time >> 11) * 3600 +
-	86400 * ((date & 31) - 1 + day_n[month] + (year / 4) + year * 365 -
+	(long long int)86400 * ((date & 31) - 1 + day_n[month] + (year / 4) + year * 365 -
 		 ((year & 3) == 0 && month < 2 ? 1 : 0) + 3653);
     /* days since 1.1.70 plus 80's leap day */
+
+    /* "time_t" is specified as "any arithmetic type" by POSIX, it can be as
+       small as a signed integer */
+    if (tempsecs > INT32_MAX) {
```
pali: This is not a correct check. Type `time_t` can be also floating-point...

Also on more systems `time_t` is 64bit therefore limiting it to just 32bit is not a good idea.

newsoft: Theoretically, you're right. In practice, no system ever used a floating-point type, and POSIX.1-2013 now requires time_t to be an integer type.

Also, time_t is 32-bit on some platforms (hence the upcoming Year 2038 bug). I agree that the check could be refined based on sizeof(time_t), but one can also hope that MS-DOS will have completely disappeared by year 2038 :)

pali:
> but one can also hope that MS-DOS will have completely disappeared by year 2038 :)

Would not happen. FAT32 is mandatory and the only supported FS for UEFI which is replacement for BIOS/DOS.

newsoft:
Interesting. It seems that UEFI is also affected by Year 2038 bug, according to:
https://github.com/tianocore/edk2/blob/18b00c8c8773e62b1e2c4a36f490996325570b67/StdLib/Include/sys/EfiCdefs.h#L116
and
https://github.com/tianocore/edk2/blob/18b00c8c8773e62b1e2c4a36f490996325570b67/StdLib/Include/utime.h#L51

What would be the proper fix in your opinion? Different code behaviours based on sizeof(time_t)?

pali:
IIRC on time_t is on some platforms signed on another unsigned. Also common is 64bit and 32bit...

It is pity that C standard nor POSIX standard does not have *_MAX macro which would contain maximal value for time_t (like for int, long int, etc...).

As dosfstools uses autoconf I would suggest to device autoconf/m4 macro which would calculate TIME_MAX value. And then use it.

pali:
Or another option, do not use time_t type at all and replace it by 64bit integer.

newsoft:
It sounds like a hack, but I could not think of a better hack, so let's do this.

Would you like me to prepare a PR, or do you do it on your own then?

newsoft:
Working on this over the week-end. I will prepare a separate PR.

newsoft:
Ok, this is not easy as it sounds.

`time(NULL)` will fit into `uint32_t` until February 7, 2106, so I guess we are good on that front. There will be a small feature gap between February 7, 2106 (maximum Epoch supported by `uint32_t`) and December 31, 2107 (maximum date supported by FAT specification), but we are bikeshedding at this point.

But how to fix the following code?
```
    time_t date;
    date = date_dos2unix(le16toh(file->dir_ent.time), le16toh(file->dir_ent.date));
    tm = localtime(&date);
```

`localtime` takes `time_t *` as an argument. Shall I just cast `uint32_t` into `time_t`? Then platforms where `time_t` is defined as signed 32-bit integer *will* be hit by the Year 2038 bug (but I guess those are hopeless anyway).

What should be the behaviour in case of truncation?
- Truncate silently ;
- Warn user ;
- Assert or fail in any other way.

pali:
> There will be a small feature gap between February 7, 2106 (maximum Epoch supported by uint32_t) and December 31, 2107 (maximum date supported by FAT specification), but we are bikeshedding at this point.

Either ignore this problem OR use uint64_t type.

> But how to fix the following code?

This is in file_stat function and its purpose is to just format passed timestamp to string which would be shown to human. There are two options: write your own formatter which would not be affected by year 2038 bug OR ignore this problem. As this function does not change how filesystem is formatted/changed, it is not so critical. It is just for human readable output and system with year 2038 bug would have bigger problems as some human readable message would not be accure.

Or if you can detect that overflow occurs, then you can throw warning OR round value... But I'm not sure if is is possible to detect if overflow occured.

Assert or failure is for sure the bad idea. Lot of time those tools are run by scripts which discards standard output, therefore fully discard this human readable message.

pali:
You can detect overflow, e.g.:
```c
uint32_t input = ...
time_t output = input;
if (output < 0 || (uint32_t)output != input) {
    /* date is not representatable */
}
```

pali:
I prepared fix for this problem in my branch year-2038, see diff: https://github.com/dosfstools/dosfstools/compare/master...pali:year-2038

pali:
@newsoft can you look at it?

pali:
I created a new pull request for Year 2038: https://github.com/dosfstools/dosfstools/pull/94

PR #68, opened 2017-11-09
Fix commit 51da71a - Handle conversion between 0xe5 and 0x05

pali:
In commit 51da71a is partial fix for handling leading 0xE5 byte. But fix is not complete, on another two places it needs to be done. In fsck.fat when converting file name entry from FAT directory and in mkfs.fat when writing label to FAT directory.

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -298,7 +298,7 @@ static int bad_name(DOS_FILE * file)
 	return 0;
 
     for (i = 0; i < MSDOS_NAME; i++) {
-	if (name[i] < ' ' || name[i] == 0x7f)
+	if ((name[i] < ' ' && !(i == 0 && name[0] == 0x05)) || name[i] == 0x7f)
 	    return 1;
 	if (name[i] > 0x7f)
 	    ++suspicious;
--- a/src/file.c
+++ b/src/file.c
@@ -66,7 +66,13 @@ char *file_name(unsigned char *fixed)
     int i, j;
 
     p = path;
-    for (i = j = 0; i < 8; i++)
+    i = j = 0;
+    if (fixed[0] == 0x05) {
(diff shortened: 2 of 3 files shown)


pali:
Fixes: https://github.com/dosfstools/dosfstools/issues/69

src/check.c
```diff
@@ -298,7 +298,7 @@ static int bad_name(DOS_FILE * file)
 	return 0;
 
     for (i = 0; i < MSDOS_NAME; i++) {
-	if (name[i] < ' ' || name[i] == 0x7f)
+	if ((name[i] < ' ' && name[i] != 0x05) || name[i] == 0x7f)
```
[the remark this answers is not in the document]
pali: Thanks for spotting this problem. Now it should be fixed.

pali:
@andreasbombe that problem should be fixed, can you review it?

src/check.c
```diff
@@ -298,9 +298,9 @@ static int bad_name(DOS_FILE * file)
 	return 0;
 
     for (i = 0; i < MSDOS_NAME; i++) {
-	if (name[i] < ' ' || name[i] == 0x7f)
+	if ((name[i] < ' ' && !(i == 0 && name[0] == 0x05)) || name[i] == 0x7f)
 	    return 1;
-	if (name[i] > 0x7f)
+	if (name[i] > 0x7f || (i == 0 && name[0] == 0x05))
```
andreasbombe: I think you added this in the revision, but what is the purpose? This will cause it to fail in Atari mode if it starts with 0x05, is that intended?

pali: Ou, right, this one is used only for atari mode. Purpose was as in above line translate first 0x05 to 0xE5.

pali: Removed.

PR #73, opened 2017-11-29
fatlabel: Fix logic for FAT label

pali:
Changes:
* Show only label stored in root directory
* Print label without padding spaces
* Clear ATTR_VOLUME from directory entry when removing label
* Optimize size of the binary
* Add section COMPATIBILITY and BUGS to fatlabel

MS-DOS 6.22, MS-DOS 7.10, Windows 98, Windows XP and also Windows 10
ignores label stored in boot sector. When label in boot sector is non-empty
and label in root directory is missing then they treat situation as there
is no label for particular disk. So when labels in boot sector and root
directory are out of sync, fatlabel shown different label as those MS-DOS
and Windows systems. To make behavior between DOS, Windows and Linux
consistent, ignores label in boot sector also in fatlabel.

See discussion for more details: https://www.spinics.net/lists/kernel/msg2645732.html


pali:
@andreasbombe PING! Would you accept these changes? Or is there needed something to change/rework?

wdlkmpx:
Maybe @andy-shev can merge this?

andy-shev:
@wdlkmpx 
I have no push access to this repo. Just helped with review.

Issue #66, opened 2017-09-28
Code page for FAT volume label


pali:
FAT volume label specified either for mkfs.fat or for fatlabel is treated as is -- as binary bytes.

But according to specification and also observed from Windows behavior, FAT volume label is interpreted according to current DOS code page. Basically in the same way as fsck.fat handle file names. As linux does not have "current DOS code page", there is no way to get one. fsck.fat has already option `-c` for specifying it.

I would propose `-c` option also for fatlabel and mkfs.fat to allow specify how should be FAT label treated when reading it or writing it.

pali:
Following request adds support for `-c` and `--codepage` option into fatlabel: https://github.com/dosfstools/dosfstools/pull/76

pali:
And https://github.com/dosfstools/dosfstools/issues/82 adds support for `--codepage` option into mkfs.fat.

PR #76, opened 2018-03-07
Codepage support for fsck.fat and fatlabel

pali:
* Initialize only LC_CTYPE locale
* Initialize iconv "from codepage" with //TRANSLIT
* fsck.fat: Change default DOS codepage to 850
* fatlabel: Add support for --codepage option
* manpages: Add section DOS CODEPAGES to fatlabel

See also: https://github.com/dosfstools/dosfstools/issues/66

### DIFF
--- a/src/charconv.c
+++ b/src/charconv.c
@@ -7,8 +7,8 @@
 static iconv_t iconv_init_codepage(int codepage)
 {
     iconv_t result;
-    char codepage_name[16];
-    snprintf(codepage_name, sizeof(codepage_name), "CP%d", codepage);
+    char codepage_name[32];
+    snprintf(codepage_name, sizeof(codepage_name), "CP%d//TRANSLIT", codepage);
     result = iconv_open(nl_langinfo(CODESET), codepage_name);
     if (result == (iconv_t) - 1)
 	perror(codepage_name);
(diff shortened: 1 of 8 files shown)


pali:
@andreasbombe can you review these changes?

src/charconv.c
```diff
@@ -29,7 +29,7 @@ static int init_conversion(int codepage)
 	initialized = 1;
 	if (codepage < 0)
 	    codepage = DEFAULT_DOS_CODEPAGE;
-	setlocale(LC_ALL, "");	/* initialize locale */
+	setlocale(LC_CTYPE, "");	/* initialize locale for CODESET */
```
andreasbombe: Is there a reason to change this? If it's just "it's not strictly needed to do `LC_ALL`" then I would prefer to leave it as it is.

Otherwise the changes look good.

pali: LC_ALL includes also LC_MESSAGES. And LC_MESSAGES is used by functions like strerror(). Therefore if you have system locale set to non-English then you will have partially translated messages, e.g.: https://github.com/dosfstools/dosfstools/blob/bfeedfce3a702a04eb8ad302d40aff623445e931/src/io.c#L149-L151

When you initialize just LC_CTYPE, then LC_MESSAGES stays in its default (C locale) and so error messages printed by strerror would be in C locale (in English) despite how is LC_ALL or LC_MESSAGES env variable set.

PR #79, opened 2018-04-18
fsck.fat: Check that . and .. are first two entries available only once

### LINKED ISSUE #32: fsck.fat does not complain when "." and ".." are not the very first non-root directory table entries
`fsck.fat` does not detect it as an error when dot and dot-dot entries are not the very first two entries in a non-root directory table; it will only complain when these entries are missing altogether. Microsoft's ScanDisk considers either condition to be an error. Additionally, even when it is detected, `fsck.fat` cannot fix the problem.

Issue found while investigating [QEMU bug #1599539](https://bugs.launchpad.net/qemu/+bug/1599539).

pali:
This change only detects problem, there is no repair support yet.

Fixes #32

### DIFF
--- a/manpages/fsck.fat.8.in
+++ b/manpages/fsck.fat.8.in
@@ -90,7 +90,9 @@ Additionally, the following problems are detected, but not repaired:
 .IP "*" 4
 Invalid parameters in boot sector
 .IP "*" 4
-Absence of . and .. entries in non\-root directories
+Absence of or too many . and .. entries in non-root directories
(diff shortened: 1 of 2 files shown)


fstirlitz:
If there is no repair support, then this change doesn't 'fix' #32.

I've got a full solution written myself, but it's quite invasive and a bit hacky: #80.

pali:
Replaced by proper solution in pull request https://github.com/dosfstools/dosfstools/pull/80

PR #81, opened 2018-05-05
fsck.fat: Fix dirty volume check

relan:
ClnShutBitMask and HrdErrBitMask bits are in the second cluster entry (FAT[1]). The first entry (FAT[0]) contains BPB_Media in its low 8 bits and all other bits are set to 1. Thus, fsck.fat never detected dirty volumes.

See https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/fatgen103.doc, page 18.

### DIFF
--- a/src/fat.c
+++ b/src/fat.c
@@ -118,8 +118,8 @@ void read_fat(DOS_FS * fs)
     }
     if (second && memcmp(first, second, eff_size) != 0) {
 	FAT_ENTRY first_media, second_media;
-	get_fat(&first_media, first, 0, fs);
-	get_fat(&second_media, second, 0, fs);
+	get_fat(&first_media, first, 1, fs);
+	get_fat(&second_media, second, 1, fs);
 	first_ok = (first_media.value & FAT_EXTD(fs)) == FAT_EXTD(fs);
 	second_ok = (second_media.value & FAT_EXTD(fs)) == FAT_EXTD(fs);
 	if (first_ok && !second_ok) {


pali:
If you try to mount FAT filesystem with dirty volume set on recent Linux kernel then kernel print to dmesg warning about dirty bit. And if you run fsck.fat to fix this problem and mount it again then kernel does not print dirty bit warning anymore.

So based on this observation, it must have been already working.

And it is strange if you say that not. So either both fsck.fat and Linux kernel use different bits for indication of dirty bit and then this patch just breaks Linux kernel detection OR some other fsck.fat code clears correct bits...

I would suggest to do tests against both Windows and Linux kernels to check if this patch would not break Linux kernel support. And also verify which bits uses Linux kernel for indication that filesystem is dirty.

relan:
I've made the following experiment:

1. Created a FAT32 FS with mkfs.fat
2. Set first two FAT entries to 0 (which should be totally invalid)
3. Mounted this image under Linux 4.9

Result: no warnings in dmesg, FS just works (files with non-zero size can be created and deleted). The image after unmout still has two first FAT entries zeroed out.

I also could not find any dirty FAT checks in Linux source code. So this change in fsck.fat should not affect Linux vfat driver.

Will check Windows later.

pali:
> I also could not find any dirty FAT checks in Linux source code.

Try again and search for string `FAT_STATE_DIRTY`.

E.g. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fat/inode.c#n653

When you try to mount FAT volume with dirty bit set, then kernel show a big warning `Volume was not properly unmounted. Some data may be corrupt. Please run fsck.` in dmesg.

relan:
> Try again and search for string FAT_STATE_DIRTY.

That's a different bit in boot sector.

This MR is about handling dirty bits in FATs.

pali:
Ou right, sorry my fault. There are really two places where dirty bit can be set. First is in boot sector and second is in reserved FAT cluster.

If we are going to support also dirty bit in reserved FAT cluster, when we should add support for it also into Linux kernel vfat/msdos driver.

Anyway, have you tested what is Windows doing with such dirty filesystem? And if chkdsk can repair it?

relan:
> If we are going to support also dirty bit in reserved FAT cluster, when we should add support for it also into Linux kernel vfat/msdos driver.

fsck.fat already supports FAT dirty bits, it just examines them in wrong position.

> Anyway, have you tested what is Windows doing with such dirty filesystem? And if chkdsk can repair it?

Nope. Will check later.

pali:
> fsck.fat already supports FAT dirty bits, it just examines them in wrong position.

So it does not support it :-)

Anyway, are you going to extend kernel vfat/msdos driver to handle also this dirty bit?

andreasbombe:
Regarding the patch in this pull request, I don't see what even the intention was. You change two `get_fat()` invocations that are supposed to get the media byte to get cluster 1 instead of cluster 0. Yet, as you yourself note, the media byte is in cluster 0.

As it stands this patch does nothing but introduce a bug.

pali:
@relan: I looked deeply at this problem and @andreasbombe is right. Patch in this pull request does nothing with dirty bit volume check nor with ClnShutBitMask bit. And really introduce a bug as it breaks media byte check.

I tried to implement setting ClnShutBit correcly, it is in clean-shutdown-bit branch, see:
https://github.com/dosfstools/dosfstools/compare/master...pali:clean-shutdown-bit

relan:
As far as I understand the spec, it's perfectly fine for all FATs except one to be dirty: software must use this indication to select the proper FAT on mount. Those bits in FATs have nothing to do with the dirty bit in the boot sector.

Anyway, I don't have any time to investigate this, sorry for the noise.

pali:
@andreasbombe can you look at my changes in clean-shutdown-bit branch? Should I create a pull request for it?

pali:
@andreasbombe can you comment?

pali:
No response for more then month, so I created pull request https://github.com/dosfstools/dosfstools/pull/100

PR #86, opened 2018-05-25
Add support for reading/restoring FAT32 backup boot sectors.

dmo2118:
Some notes:

* A few of the early boot sector tests lost their specific error messages when they got eaten by try_boot_sector() in favor of the more ambiguous "Boot sector is unusable" message.
* ~~If the backup boot sector needs to be restored, fsck.fat users are presented with the non-choice of doing the restore, or quitting. (And incidentally, get_choice() now allows only one choice.) Mostly this is so that the rest of dosfstools can continue to read/write sector 0 with the expectation that that's where the boot sector is.~~ (removed in 89826a4)
* ~~"Backup boot sector in unexpected location" might be a bit vague.~~ "Backup boot sector pointer incorrect" is less vague, but that will still be meaningless to anyone not familiar with FAT internals. It's for the probably rare circumstance where (a) the original boot sector is hosed, (b) fsck.fat found a valid backup boot sector, and (c) the location of the backup boot sector as indicated in the backup boot sector does not match the location where the backup boot sector was initially found.

I've got "Allow edits from maintainers" checked, so you know, by all means.


dmo2118:
(Don't mind me, just screwing with my repo...)

dmo2118:
(Rebased on dosfstools/master & squashed.)

src/boot.c
```diff
@@ -446,30 +454,120 @@ static void check_fat_state_bit(DOS_FS * fs, void *b)
...
+	b->fats &&
+	/* ...else number of FATs is zero. (NTFS is like this.) */
+
+	!(*byte_offset % logical_sector_size);
+	/* ...else the boot sector isn't on a logical sector boundary. */
+}
+
+static int has_signature(const void *sector)
+{
+    return le16toh(((const uint16_t *)sector)[255]) == 0xaa55;
```
pali: ... or you can directly check `sector[510] == 0xFF && sector[511] == 0xAA`, make `sector` of type `unsigned char*` (or `uint8_t*`) to prevent problems with endianity...

dmo2118: Sure, why not.

(Though I think my way would be around *a whole nanosecond* faster!)

pali:
@dmo2118: In pull request https://github.com/dosfstools/dosfstools/pull/117 was extended/fixed mkfs.fat code to properly generate backups of boot/info sectors. Could you rebase this your pull request on top of the master branch and check that code in this your pull request is compatible with those changes?

dmo2118:
It worked before, but I fixed a bounds check, and it should be a bit better now.

Tested it a bit with small reserved regions:

```
#!/bin/sh -e

dosfstools=~/src/dosfstools/src
for r in 2 3 4 5 6 7 8
do
	echo "# Reserved sectors: $r"
	rm -f fat32.img
	truncate -s 40M fat32.img
	$dosfstools/mkfs.fat -v -F 32 -s 1 -R $r fat32.img
	dd bs=512 count=2 conv=notrunc if=/dev/zero of=fat32.img
	$dosfstools/fsck.fat fat32.img || true
	hexdump -C fat32.img
done
```

2 reserved sectors: Everything's gone because there wasn't any room for a backup.
3 and 6 reserved sectors: No FSInfo backup, so that has to be rebuilt. 
Otherwise: It works.

tests/Makefile.am
```diff
@@ -38,30 +40,33 @@ TESTS = referenceFAT12.mkfs              \
 
 
 XFAIL_TESTS = check-dot_entries.fsck \
-	      check-huge.fsck
+	      check-huge.fsck \
+	      check-backup_boot_sector_mismatch.fsck
```
pali, line 44: Cannot we fix fsck to be able to repair also this `check-backup_boot_sector_mismatch.fsck` filesystem?

tests/Makefile.am
```diff
@@ -4,19 +4,21 @@ if AUTOMAKE_TEST_COMPAT
...
-	check-circular_chain.fsck        \
-	check-duplicate_names.fsck       \
-	check-dot_entries.fsck           \
-	check-huge.fsck                  \
+TESTS = referenceFAT12.mkfs                    \
+	referenceFAT16.mkfs                    \
+	referenceFAT32.mkfs                    \
+	referenceFAT32mbr.mkfs                 \
+	mkfs-fat32_2_res_sects.mkfs            \
+	check-backup_boot_sector_mismatch.fsck \
```
pali, line 12: I would suggest to rename this test file to be shorter, so it would not be needed to realign all other file names. There are lot of other pull requests and changes which may result in merge conflicts just to filename alignment.

What about renaming it to just `check-backup_boot_sec_mism.fsck`?

pali: Realign all of file names now already causes merge conflicts and therefore it is not possible to merge this PR yet.

pali:
@dmo2118: Can you look at above mentioned issues?

dmo2118:
I haven't forgotten about this, but I've been short on free weekends for quite a while now. Sorry.

tests/Makefile.am
dmo2118: Doable, I guess. I've downgraded this to a warning: If the backup boot sector was found in one place, but the backup boot sector says the backup is in a different sector, `fsck.fat` updates the pointer in the boot sector to where the backup boot sector was actually found.

Alternately, `fsck.fat` could move the backup boot sector to where the pointer says it should be, but that feels excessive.

dmo2118:
Things are less hectic for me now, FWIW.

src/boot.c
```diff
@@ -412,31 +427,131 @@ static void read_fsinfo(DOS_FS * fs, struct boot_sector *b, unsigned int lss)
...
+		    die("Backup boot sector not found.");
+		if (sector != 6 &&
+		    sector != 12 &&
+		    sector != 24 &&
+		    try_backup_boot_sector(&boot_sector, sector, &b)) {
+		    break;
+		}
+		sector++;
+	    }
+	}
```
pali, line 523: I do not like this automatic detection. Reason is that this is heuristic and does not have to be accurate. And so it may cause more issues when restoring broken filesystem.

So at least some interactive confirmation from user is required, to prevent some future damage.

You can look for inspiration in code https://github.com/dosfstools/dosfstools/pull/155 which added support for fixing first fat cluster. There is special command line option which can be used by user to instruct fsck.fat which FAT should use for repairing (if FATs are different). So something similar could be useful for specifying from which sector should be read backup boot sector.

Also I would suggest to implement more strict checking for backup boot sector (only backup). Look for example into util-linux project where should be check for boot sector very similar / same as is doing linux kernel during mounting fs: https://github.com/karelzak/util-linux/blob/stable/v2.37/libblkid/src/superblocks/vfat.c#L184-L265

src/boot.c
```diff
@@ -294,6 +295,20 @@ static void check_backup_boot(DOS_FS * fs, struct boot_sector *b, unsigned int l
 	    return;
     }
 
+    if (boot_sector && fs->backupboot_start &&
```
pali, line 298: Is this check correct? Because by default is boot_sector zero.

pali (changes requested):
Hello! Sorry for waiting longer time. I had to look at this code more properly as it was not obvious for me what it is doing. And from past pull requests and issues I know that backup boot sector and info sector topics are complicated in FAT. Below are some review comments.

src/boot.c
```diff
@@ -412,31 +427,131 @@ static void read_fsinfo(DOS_FS * fs, struct boot_sector *b, unsigned int lss)
...
+		    sector != 24 &&
+		    try_backup_boot_sector(&boot_sector, sector, &b)) {
+		    break;
+		}
+		sector++;
+	    }
+	}
+    }
+
     logical_sector_size = GET_UNALIGNED_W(b.sector_size);
```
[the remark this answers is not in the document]
dmo2118: That check is in `try_boot_sector`, the `byte_offset % logical_sector_size` bit.

The `offset` parameter in `try_(backup_)boot_sector` is a multiple of 512 (`SECTOR_SIZE`), because any other sector size will be a multiple of that. For example, with a logical sector size of 4 KB, it might find a backup boot sector at 48 * 512, then right after it'll do a `GET_UNALIGNED_W(b.sector_size)`, and `boot_sector` will be set to 48 * 512 / 4096 == 6.

src/boot.c
```diff
@@ -412,31 +427,131 @@ static void read_fsinfo(DOS_FS * fs, struct boot_sector *b, unsigned int lss)
...
+    if (boot_sector) {
+	printf("Restoring boot sector from backup found at logical sector %u.\n",
+	       (unsigned)boot_sector);
+
+	/* The backup boot sector in FAT32 *may* be followed by up to two more
+	 * backup sectors, usually FSInfo, followed by more boot code. Backup
+	 * sectors should be tagged with a 55 AA signature. */
+	void *bdata = alloc(logical_sector_size);
+	unsigned i;
+	for (i = 0; i != 3; ++i) {
```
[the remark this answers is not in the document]
dmo2118: Well, it's **up to** three sectors, stopping if one of them doesn't have a `55 aa` signature; `try_backup_boot_sector` calls `has_signature` for that.

fatgen103.pdf just says three sectors. I know `mkfs.fat` uses less, but is there really anything out there that uses more than three sectors here? It's certainly doable to not limit the backup boot record to three sectors, but I start worrying about what could theoretically get accidentally overwritten if we do that.

---

I'm operating under the assumption that boot code can only be restored on a best-effort basis.

Windows 95/98 had a three sector boot record as fatgen103.pdf implied, with each sector tagged with `55 aa`:
- Sector 0: BPB/boot sector/boot code
- Sector 1: FSInfo, mostly empty
- Sector 2: More boot code
- Sectors 6-8: Backup of sectors 0-2

But when we get to Windows 2000 and later: ([hex dump](https://thestarman.pcministry.com/asm/mbr/FAT32xp7comp.htm); each of these still has a `55 aa`)
- Sector 0: BPB/boot sector/boot code
- Sector 1: FSInfo, mostly empty
- Sector 2: Empty, except for a `55 aa` signature
- Sectors 6-8: Backup of sectors 0-2
- Sector 12: More boot code. **There's no backup of this.**

...So hopefully if something burns sector 0, but leaves sector 6, then sector 12 will be left alone as well, and restoring the backup boot record makes the drive bootable.

src/boot.c
dmo2118: Fair enough, I've added an interactive prompt, with a subset of the info from `-v`. My thinking here is that interactive use needs less checking than something fully automated.

I think the test is sufficiently permissive here that no potentially valid boot sector would be missed by `fsck.fat` during the interactive search, so nobody would need to force a specific backup sector on the command line.

(Will probably tweak this later.)

src/boot.c
dmo2118: Yeah, that's only if the backup boot sector is being used (i.e. `boot_sector != 0`).

pali: So what is this code doing? And why is this check needed? Because it is completely unclear / unreadable what is going here...

src/boot.c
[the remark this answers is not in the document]
pali: Ok.

src/boot.c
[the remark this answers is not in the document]
pali: > Well, it's **up to** three sectors

No, it is not _up to_. Please look at link which I have already wrote: https://github.com/dosfstools/dosfstools/pull/117#issuecomment-573360348

> fatgen103.pdf just says three sectors.

No, please really look at details in https://github.com/dosfstools/dosfstools/pull/117#issuecomment-573360348
as 3 is just _example_ value.

> I know `mkfs.fat` uses less, but is there really anything out there that uses more than three sectors here?

mkfs.fat has `-R` option which allows usage more of them.

src/boot.c
pali: Usage of (backup) boot sector which does not pass above util-linux check does not make sense as it is unusable. So really fat boot sector check should be same as in kernel / util-linux.

src/boot.c
```diff
@@ -412,31 +427,152 @@ static void read_fsinfo(DOS_FS * fs, struct boot_sector *b, unsigned int lss)
...
+
+	/* Backup boot sector is at sector 6 by default. Check that first, for
+	 * valid sector sizes of 512 to 4096 bytes.
+	 *
+	 * Might not be a bad idea to only check this for disks large enough to
+	 * accommodate FAT32. */
+	if (!try_backup_boot_sector(&boot_sector, 6, &b) &&
+	    !try_backup_boot_sector(&boot_sector, 12, &b) &&
+	    !try_backup_boot_sector(&boot_sector, 24, &b) &&
+	    !try_backup_boot_sector(&boot_sector, 48, &b)) {
```
pali, line 531: Due to how you check for sector size, this code does not do what is written in comment above. Third test also checks for sector 24 with logical sector size 512 bytes. But instead third test should check only for sector 6 with logical sector size 2048 bytes.

src/boot.c
```diff
@@ -412,31 +427,152 @@ static void read_fsinfo(DOS_FS * fs, struct boot_sector *b, unsigned int lss)
...
+
+	!b->cluster_size ||
+	/* Cluster size is zero. */
+
+	!b->fats ||
+	/* Number of FATs is zero. (NTFS is like this.) */
+
+	byte_offset % logical_sector_size
+	/* ...else the boot sector isn't on a logical sector boundary. */
+	)
```
pali, line 462: As written in another comment https://github.com/dosfstools/dosfstools/pull/86#discussion_r692547126, this check is incomplete. It allows to use e.g. backup boot sector which is outside of the reserved area space, which is obviously invalid and would not work. So proper check for fat boot sector needs to be done.

dmo2118:
I'm closing this PR because I no longer have time to work on this.

PR #87, opened 2018-05-27
Add automatic code page determination, based on $LC_ALL/$LC_CTYPE/$LANG.

dmo2118:
...Because I'm lazy, and I don't want to have to always be typing `-c 437`.

Semi-related observation: the [Windows OEM code pages](https://web.archive.org/web/20171015144140/https://www.microsoft.com/resources/msdn/goglobal/default.mspx) don't entirely match the [DOS code pages](https://web.archive.org/web/20140913235232/http://support.microsoft.com/kb/117850).

And for the record, [here's the script](https://gist.github.com/dmo2118/951790d439900da2819a9326c2ad20e9) that I used to generate the code page mappings.

### DIFF
--- a/manpages/fatlabel.8.in
+++ b/manpages/fatlabel.8.in
@@ -49,7 +49,7 @@ Switch to volume ID mode.
 Remove label in label mode or generate new ID in volume ID mode.
 .IP "\fB-c\fR \fIPAGE\fR, \fB\-\-codepage\fR=\fIPAGE\fR" 4
 Use DOS codepage \fIPAGE\fR to encode/decode label.
-By default codepage 850 is used.
+By default a codepage appropriate to your locale is used.
 .IP "\fB\-h\fR, \fB\-\-help\fR" 4
 Display a help message and terminate.
 .IP "\fB\-V\fR, \fB\-\-version\fR" 4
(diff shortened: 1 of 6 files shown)


pali:
This would cause problems also on same machine. External (GUI) tools lot of times run filesystem utils without env variables, or with neutral C locale. So using fatlabel from exterernal/GUI application would lead to diffetent label as calling fatlabel manually from command line -- even on same machine. This is really not a good :-(

dmo2118:
> Windows implementation of fastfat driver uses OEM �code page for label for sure.

Yeah, but I'm thinking more along the lines of mismatches like Arabic (DOS Country.txt: CP864, Windows OEM: CP720) or Korean (CP934 vs. CP949) or en-AU (CP437 vs. CP850). It's just something I noticed that seemed a little odd. (Plus, it does say DEFAULT_DOS_CODEPAGE and not, say, DEFAULT_OEM_CODEPAGE...)

> For fatlabel I chose CP850 because same default is also in mlabel command line program and in most Windows regional settings.

Well, sure CP850 is pretty common, but there's still a huge number of people that use something else: zh-CN: CP936, en-IN: CP437, en-US: CP437, etc. :man_shrugging:

> This would cause problems also on same machine. External (GUI) tools lot of times run filesystem utils without env variables, or with neutral C locale. So using fatlabel from exterernal/GUI application would lead to diffetent label as calling fatlabel manually from command line -- even on same machine. This is really not a good :-(

Okay, yeah, that could be trouble. `su(1)` toasts `$LANG` on my end.

pali:
> Yeah, but I'm thinking more along the lines of mismatches like Arabic (DOS Country.txt: CP864, Windows OEM: CP720) or Korean (CP934 vs. CP949) or en-AU (CP437 vs. CP850).

This is Microsoft mess in Windows... nothing which can be fixed or changed on Linux nor in dosfstools. Or maybe... there can be also exceptions in Windows code or that documentation is wrong... So to verify if these are really mistakes, it is needed to do tests on Windows and DOS machines.

> Well, sure CP850 is pretty common, but there's still a huge number of people that use something else: zh-CN: CP936, en-IN: CP437, en-US: CP437, etc. :man_shrugging:

Yes, I'm in region where CP852 is used.

PR #92, opened 2018-08-12
mkfs.fat: Fix parsing and validation of numeric values

pali:
Properly handle errors from strtoul() function. To detect overflow/underflows it is needed to check for errno. Also there is a need to check special case if argument is empty string.

### DIFF
--- a/src/fatlabel.c
+++ b/src/fatlabel.c
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <getopt.h>
+#include <ctype.h>
 
 #include "common.h"
 #include "fsck.fat.h"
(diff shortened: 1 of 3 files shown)


src/fatlabel.c
```diff
@@ -116,14 +116,14 @@ static void handle_volid(bool change, bool reset, const char *device, const char
...
+    long long conversion;
     uint32_t serial = 0;
 
     if (change) {
 	errno = 0;
-	conversion = strtoull(newserial, &tmp, 16);
+	conversion = strtoll(newserial, &tmp, 16);
 
-	if (!*newserial || *tmp) {
+	if (!*newserial || isspace(*newserial) || *tmp || conversion < 0) {
```
lu-zero: why you aren't checking for errno here as well? probably would make sense to factor the checks so they stay consistent across the codebase.

pali: Because it is already checked few lined below, see:
https://github.com/dosfstools/dosfstools/pull/92/files#diff-648f9bb939c2d42dafca95e979510a64L134

lu-zero: I see, alla in all the patch looks fine even if the checks could be more uniform, but that could be reworked later I guess.

pali: ok, thanks for review.

PR #93, opened 2018-08-14
fsck.fat: Check and fix label

pali:
Checks the volume label from the root directory entry that is valid and
matches the label stored in boot sector.

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -4,6 +4,7 @@
    Copyright (C) 1998 Roman Hodek <Roman.Hodek@informatik.uni-erlangen.de>
    Copyright (C) 2008-2014 Daniel Baumann <mail@daniel-baumann.ch>
    Copyright (C) 2015 Andreas Bombe <aeb@debian.org>
+   Copyright (C) 2017-2021 Pali Rohár <pali.rohar@gmail.com>
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -29,6 +30,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <errno.h>
+#include <ctype.h>
+#include <wctype.h>
 
 #include "common.h"
 #include "fsck.fat.h"
@@ -37,6 +41,8 @@
 #include "file.h"
 #include "lfn.h"
 #include "check.h"
+#include "boot.h"
+#include "charconv.h"
 
 
 /* the longest path on the filesystem that can be handled by path_name() */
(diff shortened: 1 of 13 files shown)


andreasbombe:
I think that fresh Windows generated FAT file systems shouldn't make `fsck.fat` report errors, which is what the check finding an existing root directory label and empty boot sector label does if I'm not mistaken.

Generally I feel that these questions about "root directory" and "boot sector" labels are too confusing for users who don't have in depth knowledge of FAT file systems. The questions should be stripped down to only actual errors (both labels exist and don't match) and reformulated, but the reformulation is something I can do later.

I'd be more comfortable with phasing out the use of boot sector labels so probably the default in case of mismatch should be to clear the boot sector label. Relatedly, maybe `mkfs.fat` shouldn't write the boot sector label anymore (or it should stop after the next release).

pali:
Reason for having synchronized boot label and volume label is compatibility with older dosfstools, older blkid (and libblkid library), plus also to be "more" compliant with Microsoft FAT specification.

Lot of distributions still use older blkid which read label also from boot sector. And also all distributions use dosfstools which fatlabel still read/write boot label (usage of volume label from root directory is only in git, unreleased fatlabel version). And also I saw usage of older dosfstools 2.x just because it is licensed under GPLv2 and not GPLv3 (some companies have problems with GPLv3 and software under this license is not allowed, but GPLv2 is OK).

So the whole point is that these changes make fsck.fat to "repair" FAT filesystem to state, to be compatible between different Linux distributions which use different versions of libblkid and dosfstools, and also with Windows.

src/check.c
```diff
@@ -995,3 +1000,209 @@ int scan_root(DOS_FS * fs)
...
+            remove_label(fs);
+            offset = 0;
+            break;
+        }
+    }
+
+    if (offset != 0 && memcmp(fs->label, "NO NAME    ", 11) != 0 && memcmp(fs->label, doslabel, 11) != 0) {
+        memcpy(buffer, pretty_label(doslabel), sizeof(buffer));
+        printf("Volume label '%s' stored in root directory and label '%s' stored in boot sector and different.\n", buffer, pretty_label(fs->label));
+        switch (get_choice(1, "  Auto-coping volume label from root directory to boot sector.",
```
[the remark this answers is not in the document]
pali: Fixed.

pali:
Also note that based on my October 2017 testing (https://www.spinics.net/lists/kernel/msg2640891.html), MS-DOS format/label tools, Win98 format/label tools and also Linux mformat/mlabel tools write label to *both* locations (boot sector and root directory).

metux:
IMHO, both patches should be squashed into one.

pali:
Yes! All fixups would be squashed at merge time.

src/check.c
```diff
@@ -995,3 +1000,209 @@ int scan_root(DOS_FS * fs)
...
+                break;
+            }
+        }
+    }
+
+again:
+
+    if (offset != 0 && memcmp(fs->label, "NO NAME    ", 11) == 0 && memcmp(doslabel, "NO NAME    ", 11) != 0) {
+        printf("There is no label in boot sector, but there is volume label '%s' stored in root directory\n", pretty_label(doslabel));
+        switch (get_choice(1, "  Auto-coping volume label from root directory to boot sector.",
```
fstirlitz: Also, 'copying'.

pabs3:
The common.c fix also applies to git master, so please move it there instead of having it added by this PR.

The two fixup commits to check.c should get merged into the first commit that makes the change, before this PR is merged IMO.

pali:
I wrote that I would squash all fixups before merging. Github is lot of times broken when you do force push, so for reviewing code on github it is better for trying to eliminate usage of force push, just fixups. And use force push (with squashing all changes) after code is reviewed and ready to merge.

src/check.c
```diff
@@ -995,3 +1000,210 @@ int scan_root(DOS_FS * fs)
...
+        memcpy(doslabel, de.name, 11);
+        if (doslabel[0] == 0x05)
+            doslabel[0] = 0xe5;
+        ret = validate_volume_label(NULL, (unsigned char *)doslabel);
+        if (ret & ~0x1) {
+            printf("Volume label '%s' stored in root directory is not valid.\n", pretty_label(doslabel));
+            switch (get_choice(1, "  Auto-removing label.",
+                               2,
+                               1, "Remove invalid label",
+                               2, "Set new label")) {
```
fstirlitz: This also should have an ‘ignore’ option.

I often set up labels of fixed file systems to be the same as their respective mount points for easier identification (e.g. in GRUB `ls -l`, or to have convenient handles in `/dev/disk/by-label`; `e2fsck` even displays the label instead of the device node’s pathname when checking the file system); this includes setting the label of the EFI system partition to `/boot/efi`. While it may be invalid according to the FAT specification, the violation is pretty benign. Even Microsoft’s checking tools don’t consistently detect it: the DOS version of ScanDisk in Windows 98 ignores it, although the GUI version replaces `/` with `-` and converts the label to uppercase. (DOS utilities mostly ignoring volume labels is probably one reason why Microsoft chose the ‘volume label’ attribute bit for LFN entries, after all.) Otherwise most software only reads the label to display it to the user.

pali: Hm... Current fatlabel and mkfs.fat code in git does not allow you to set such label (as they are invalid according to specification). But if you are saying that such labels works fine, and more important that there is also usecase for it, I would this about it a bit more... What about adding a new command line option for these tools which cause that tools would not treat such labels as invalid? I do not think it should be a default behavior of dosfstools to create a filesystem which is invalid according to specification, but if there is reasonable argument for such behavior, additional option could allow usage of it.

fstirlitz: I use version 4.1 (2017-01-24) from Debian repositories, which still allows it. And I wouldn’t mind it hidden behind a command-line switch, although can be hard to pass command-line flags to boot-time fsck. Maybe an environment variable would work better.

pali: I think that for full label check & fix support would be better to first have PR https://github.com/dosfstools/dosfstools/pull/133 merged and this PR rebased on top of them. This label check & fix support started to be more complicated.

pali:
Now after merging other pull requests, I rebased this pull request on top of the master branch and included new changes into this pull request.

pali:
Merged now.

PR #94, opened 2018-08-17
Year 2038 problem

pali:
**fsck.fat**: Do not use `time_t` type and `strftime()` function which are affected by the Year 2038 Bug. Instead parse date/time directly from DOS format which avoids conversion from DOS to UNIX + conversion from UNIX to string.

**mkfs.fat**: Check return values of `time()`, `gmtime()` and `localtime()` function which may fail. Fallback to date/time 1.1.1980 00:00:00 when those functions fail. For generating volume id, add fallback code based on `rand()` function.

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -220,7 +220,7 @@ static char *device_name = NULL;	/* Name of the device on which to create the fi
 static int check = FALSE;	/* Default to no readablity checking */
 static int verbose = 0;		/* Default to verbose mode off */
 static long volume_id;		/* Volume ID number */
-static time_t create_time;	/* Creation time */
+static time_t create_time = -1;	/* Creation time */
 static char *volume_name = initial_volume_name;	/* Volume name */
 static uint64_t blocks;	/* Number of blocks in filesystem */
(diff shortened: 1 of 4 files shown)


newsoft:
Sorry, I was on vacation with no Internet access, but SGTM. Thanks for fixing this.

src/mkfs.fat.c
```diff
@@ -1170,17 +1170,26 @@ static void setup_tables(void)
 	if (de->name[0] == 0xe5)
 	    de->name[0] = 0x05;
 	de->attr = ATTR_VOLUME;
-	if (!invariant)
+	if (create_time != (time_t)-1) {
```
bjornfor: create_time is initialized to -1 (and is of type time_t), so what is the purpose of this check? It will never be true, right?

pali: See: https://github.com/dosfstools/dosfstools/pull/94/files#diff-47b7a74c103277bcad9443e559f1ee9bR1375

`create_time` is initialized to `-1`. And later is updated to up-to-date value, but only in case system is not affected by Year 2038 bug.

PR #97, opened 2018-08-29
mkfs.fat: Avoid usage of undefined behavior in open()

pali:
Behavior of O_EXCL in open() is undefined if it is used without O_CREAT.
The only exception on Linux is when pathname is block device. In that case
when block device is already mounted, open() fails with error EBUSY.

This change avoids calling O_EXCL on path names which are not block
devices. First it opens path name normally, check file descriptor that it
is block device, and then re-open file descriptor with O_EXCL via /proc.
This is fully race-free trick without any undefined behavior in usage.

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -1343,6 +1343,9 @@ int main(int argc, char **argv)
     int create = 0;
     uint64_t cblocks = 0;
     int blocks_specified = 0;
+    int new_dev;
+    char proc_path[30];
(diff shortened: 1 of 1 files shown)


pali:
This code for double opening via /proc is causing issues for some block devices. E.g. pktcdvd mapper which does not allow to hold open fd with O_EXCL mode if there is any other fd open (even in O_RDONLY). So closing this pull request.

Issue #38, opened 2016-10-27
Usage of "Dirty Bit" does not meet FAT specification


smklein:
It seems that fsck is using byte "0x41" (FAT32) or "0x25" (FAT16) as as the containing byte of this bit.

However, the FAT specification indicates that the first two clusters of a FAT are reserved for information which includes this dirty bit (notably, not within the boot sector). It also mentions that byte 0x41/0x25 is "BS_Reserved1", which "is "Reserved (used by Windows NT). Code that formats FAT volumes should always set this byte to 0."

Apparently, fsck's behavior isn't wrong -- there are other (informal) third party sources that confirm the usage of byte 0x41 / 0x25 as dirty bits. However, I can't find any formal documentation for this "out of spec" dirty bit -- or any reasoning why the documented dirty bit is going unused.

Additionally, there doesn't seem to be any attempt by fsck to set the formally specified "ClnShutBitMask" to "clean" after a verification pass.

jristz:
Do you can link the specification document to give a check?

I think maybe as you say 0x41/0x25 is "BS_Reserved1", which is "Reserved (used by Windows NT)" maybe Windows NT add they dirty bit too aditionally to the one in the specification, so they are documenting FAT dirty bit but not NT doings.

smklein:
https://staff.washington.edu/dittrich/misc/fatgen103.pdf

Page 18:

**What are the two reserved clusters at the start of the FAT for? The first reserved cluster, FAT[0],
contains the BPB_Media byte value in its low 8 bits, and all other bits are set to 1. For example, if the
BPB_Media value is 0xF8, for FAT12 FAT[0] = 0x0FF8, for FAT16 FAT[0] = 0xFFF8, and for
FAT32 FAT[0] = 0x0FFFFFF8. The second reserved cluster, FAT[1], is set by FORMAT to the EOC
mark. On FAT12 volumes, it is not used and is simply always contains an EOC mark. For FAT16 and
FAT32, the file system driver may use the high two bits of the FAT[1] entry for dirty volume flags (all
other bits, are always left set to 1). Note that the bit location is different for FAT16 and FAT32,
because they are the high 2 bits of the entry.**

pali:
> I can't find any formal documentation for this "out of spec" dirty bit ...

[...]

PR #100, opened 2018-09-27
fsck.fat: Check for DOS Clean Shutdown bit

pali:
DOS Clean Shutdown bit in first reserved FAT entry is cleared when DOS or
Windows FAT driver mounts a volume and set is back when doing unmount.
Therefore set this bit when clearing FAT dirty bit in boot sector.

Fixes: #38

### DIFF
--- a/src/boot.c
+++ b/src/boot.c
@@ -422,26 +422,55 @@ static char print_fat_dirty_state(void)
 		      2, "No action");
 }
 
-static void check_fat_state_bit(DOS_FS * fs, void *b)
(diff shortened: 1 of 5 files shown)


jnuyens:
Please merge this important missing feature - it has been 5 months now. Downstream users need this as Windows 10 no longer sets and removes this bit, but still detects it and presents users with a 'please repair' dialog which fsck.vfat can't even repair... 

If you don't have time enough to manage this project, hand it over to someone who contributes a lot; in good OpenSource fashion. Thanks!

PR #102, opened 2018-10-21
Clarify unit of BLOCK-COUNT in mkfs.fat manpage

pali:
BLOCK-COUNT is always in KiB unit independently of the disk sector size,
FAT sector size or FAT cluster size.


fstirlitz:
‘kB’ is the metric kilobyte, i.e. 1000 bytes. You're thinking of ‘KiB’.

pali:
Ok, I updated it.

Issue #29, opened 2016-05-17
mkfs.fat manpage: Misleading description of the -h option


harrison3000:
The current description for the -h option may be interpreted as if the hidden sectors are wasted space on the filesystem, when in fact the value set by the -h option is just the partition offset...
I don't even know why they called this value "hidden sectors" in the first place, but I think the man page should clarify that it is actually the partition offset and that it will be handled automatically by mkfs.fat

fstirlitz:
More specifically, it is indeed the partition offset, but it's counted from the start of the disk only for primary partitions. For extended partitions, it's from the beginning of the cylinder where the partition starts.

Don't look at me like that. It really is that insane.

pali:
@harrison3000 and @fstirlitz what about following manpage update?
```diff
diff --git a/manpages/mkfs.fat.8.in b/manpages/mkfs.fat.8.in
index 7110732..a9cf4e9 100644
--- a/manpages/mkfs.fat.8.in
+++ b/manpages/mkfs.fat.8.in
@@ -119,9 +119,11 @@ disk geometry of \fIDEVICE\fR. They are stored into FAT boot sector. Number
 By default disk geometry is read from \fIDEVICE\fR itself. If is not available
 then LBA-Assist Translation based on total number of sectors is used.
 .IP "\fB\-h\fR \fINUMBER-OF-HIDDEN-SECTORS\fR" 4
-Select the number of hidden sectors in the volume.
+Specify the number of hidden sectors. This number represents sector which is
+beginning of the \fIDEVICE\fR partition and is stored into FAT boot sector.
 Apparently some digital cameras get indigestion if you feed them a CF card
-without such hidden sectors, this option allows you to satisfy them.
+without hidden sectors value in FAT boot sector, this option allows you to
+satisfy them. By default this value is read from the \fIDEVICE\fR partition.
 .IP "\fB\-i\fR \fIVOLUME-ID\fR" 4
 Sets the volume ID of the newly created filesystem; \fIVOLUME-ID\fR is a 32-bit
 hexadecimal number (for example, 2e24ec82).
```
Or do you have better idea how to describe it in manpage?

fstirlitz:
Maybe rather like:

> Specify the number of so-called "hidden sectors", i.e. the offset of the partition containing the file system (usually relative to disk start), as stored in the FAT boot sector. If the *DEVICE* is a partition block device, this value is determined automatically by default.

(Not sure if the camera bits are still relevant.)

[...]

PR #105, opened 2018-11-24
mkfs.fat: Remove dependency on systemd/udev

pali:
There is no need that mkfs.fat depends on systemd. It uses it only for
reading additional information about block device.

This patch replace systemd/udev dependency code by reading those additional
information about block device directly from sysfs.

So no functionality of mkfs.fat is lost or removed.

### DIFF
--- a/README
+++ b/README
@@ -5,10 +5,6 @@ under the GNU GPL version 3 or later. See the file COPYING for details.
 
 ### Build Requirements
 
-dosfstools recommends libudev. It is used in mkfs.fat to collect additional
-information about the device to format in order to refuse potentially unsafe
-operations without additional confirmation.
-
 The test suite requires the tool xxd (available as part of the vim
 distribution).
 
--- a/README.md
+++ b/README.md
@@ -5,10 +5,6 @@ under the GNU GPL version 3 or later. See the file `COPYING` for details.
 
 ### Build Requirements
 
-dosfstools recommends libudev. It is used in mkfs.fat to collect additional
-information about the device to format in order to refuse potentially unsafe
-operations without additional confirmation.
-
 The test suite requires the tool `xxd` (available as part of the `vim`
 distribution).
 
--- a/configure.ac
+++ b/configure.ac
@@ -54,6 +54,7 @@ AC_CHECK_HEADERS([\
 		  linux/fd.h \
 		  linux/hdreg.h \
 		  linux/version.h \
+		  linux/loop.h \
 		  sys/disk.h \
 		  sys/disklabel.h \
 		  sys/ioccom.h \
@@ -67,13 +68,6 @@ AC_CHECK_FUNCS([vasprintf])
 
 AC_CHECK_DECLS([getmntent], [], [], [[#include <mntent.h>]])
 AC_CHECK_DECLS([getmntinfo], [], [], [[#include <sys/mount.h>]])
-
-AC_ARG_WITH([udev], AS_HELP_STRING([--without-udev], [build without libudev support]))
-if test "x$with_udev" != "xno"; then
-  PKG_CHECK_MODULES([UDEV], [libudev],
-		  [AC_DEFINE([HAVE_UDEV], [1])],
-		  [true])
-fi
 
 # optional iconv support
 AC_ARG_WITH([iconv], AS_HELP_STRING([--without-iconv], [build without iconv support]))
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -39,13 +39,12 @@ devinfo_common_sources = device_info.c device_info.h \
 mkfs_fat_SOURCES  = mkfs.fat.c msdos_fs.h common.c common.h endian_compat.h \
 		    $(charconv_common_sources) $(devinfo_common_sources)
 mkfs_fat_CPPFLAGS = -I$(srcdir)/blkdev
-mkfs_fat_CFLAGS   = $(AM_CFLAGS) $(UDEV_CFLAGS)
-mkfs_fat_LDADD    = $(UDEV_LIBS) $(charconv_common_ldadd)
+mkfs_fat_CFLAGS   = $(AM_CFLAGS)
+mkfs_fat_LDADD    = $(charconv_common_ldadd)
 
 testdevinfo_SOURCES  = testdevinfo.c $(devinfo_common_sources)
 testdevinfo_CPPFLAGS = -I$(srcdir)/blkdev
-testdevinfo_CFLAGS   = $(AM_CFLAGS) $(UDEV_CFLAGS)
-testdevinfo_LDADD    = $(UDEV_LIBS)
+testdevinfo_CFLAGS   = $(AM_CFLAGS)
 
 
 if COMPAT_SYMLINKS
--- a/src/device_info.c
+++ b/src/device_info.c
@@ -1,6 +1,7 @@
 /* device_info.c - Collect device information for mkfs.fat
 
    Copyright (C) 2015 Andreas Bombe <aeb@debian.org>
+   Copyright (C) 2018 Pali Rohár <pali.rohar@gmail.com>
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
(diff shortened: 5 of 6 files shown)


andreasbombe:
The dependency is already optional and can be disabled even when available. Whether manually digging around sysfs is preferable needs a longer look, this looks complicated.

pali:
If you look at this change, +155 lines are added and −174 removed. Therefore implementation via udev/systemd has same complexity as direct implementation via sysfs. So there is no other value for depending on udev/systemd.

Anyway, if you think that this change is complicated, just ask for lines or block of code and I can explain it deeper if needed. I wrote similar code already for another fs tool: mkudffs.

pthfdr-42:
I definitively do not want such essential programs to depend on Poetterware.

wdlkmpx:
Honestly, this is the right approach.

dosfstools may be used in embebbed systems or something. Initramfs, etc.

With this change It's also possible to produce a more complete static/portable dosfstools..

pali:
@andreasbombe: Can you comment on changes in this pull request? Could it be merged or are there some problems with it?

fstirlitz:
Well, bad news: if `mkfs.fat` is to handle extended partitions like I described in https://github.com/dosfstools/dosfstools/issues/29#issuecomment-274871809 then extended partitions will have to be detected in some way, and the best one seems to be either to retain this dependency or to bring back `libblkid`. (Unless you want to parse the MBR yourself, but then you'll still have to detect the parent device somehow.)

And besides, I'm not Lennart's biggest fan myself, but removing things just to spite him seems a bit silly.

pali:
> Well, bad news: if `mkfs.fat` is to handle extended partitions like I described in [#29 (comment)](https://github.com/dosfstools/dosfstools/issues/29#issuecomment-274871809) then extended partitions will have to be detected in some way

Detection of extended partition is simple. Partition number needs to be bigger then 4. Kernel allocates partitions numbers 1-4 for primary paritions, so if you have one primary and one extended, then parition numbers would be one and five.

And in this pull request is already code which reads partition number of supplied device.

> and the best one seems to be either to retain this dependency or to bring back `libblkid`. (Unless you want to parse the MBR yourself, but then you'll still have to detect the parent device somehow.)
> 
> And besides, I'm not Lennart's biggest fan myself, but removing things just to spite him seems a bit silly.

There are reasons why I'm proposing this change:

1. This dependency is optional and without it all such features (like detection of partition number or if device is removable or not) are not available. After my change, these features are always available, even for statically linked build.

2. Rewriting this code without external dependency does not remove any feature and has less line of codes, see diff stat: +157 −176.

So I do not see reason why to use such external dependency.

wdlkmpx:
Pali should merge this. I think there should be at least 5 people
testing the patches in the pull requests and providing feedback

Then when it's merged more people can test it thoroughly... and in
fact the only way to see if something is broken is by merging code and
forcing people to test it [it can be reverted if it's really
broken]... as long as you don't produce a release tarball, it's  safe
to merge this.

fstirlitz:
> Detection of extended partition is simple. Partition number needs to be bigger then 4.

Assuming the partitioning scheme was MBR in the first place; and it's not enough to check if the partition was extended, it's also necessary to check if the MBR type ID is 0x05 or 0x0f, because this special case does not apply to the latter (although maybe in fact I've got this condition wrong; I'm yet to test all possible combinations of OSes, FAT variants and partition IDs).

> There are reasons why I'm proposing this change:

Fair enough, those are quite legitimate reasons. I'm only mentioning The Lennart because that reason has been brought up here by someone else and apparently you're sympathetic to it (judging by the thumbs up). Although if you later end up reimplementing substantial amount of functionality already present in those libraries, this may turn out not to be such a great idea after all.

pali:
> Assuming the partitioning scheme was MBR in the first place

The whole thing is about filling correct value into `hidden sectors` field. And this field is needed only for DOS systems which supports only MBR.

So if GPT is used I think it we could ignore this problem...

> it's also necessary to check if the MBR type ID is 0x05 or 0x0f

Linux kernel does not create partition for MBR ids 0x05, 0x0F and 0x85. So user cannot format partition with these types. These ids are reserved for creating extended partitions and Linux kernel automatically parse them exports only logical partitions which are as part of extended.

So this is not a problem too.

If you have a disk with MBR, one primary and one logical partition it would look like: /dev/sda, /dev/sda1 (primary), /dev/sda5 (logical). There is no /dev/sda4 which should represent extended partition on which are all logical partitions (/dev/sda5) stored.

fstirlitz:
> So if GPT is used I think it we could ignore this problem...

Even if the disk is GPT-partitioned, the partition may be additionally exposed in a hybrid MBR setup as a primary MBR partition. So you can't just assume partition number ≥ 5 means apply the workaround.

> Linux kernel does not create partition for MBR ids 0x05, 0x0F and 0x85. So user cannot format partition with these types. These ids are reserved for creating extended partitions and Linux kernel automatically parse them exports only logical partitions which are as part of extended.

I was referring to the ID of the extended partition containing the logical volume being formatted, obviously. I imagine the ultimate reason DOS 7.x chooses one offset or the other is that at some point the DOS kernel chooses between LBA and CHS drivers to access the partition, which have different data structures, and this is triggered by some combination of partition IDs for the EBR and for the logical volume itself that is yet to be fully investigated.

(And technically, at least on my machine, Linux does create partitions for types `0x05` and `0x0f,` although they are just stubs that only contain the EBR and, if I recall correctly, some padding sectors.)

pali:
> Even if the disk is GPT-partitioned, the partition may be additionally exposed in a hybrid MBR setup as a primary MBR partition. So you can't just assume partition number ≥ 5 means apply the workaround.

Ok, but systemd nor blkid does not help us with detection of such setup. For hybrid MBR setup you need to parse both MBR and GPT sectors manually.

And about MBR ids and offset of partitions (number of hidden sectors), are you going to fully finish your investigation how it works?

But currently it looks like that systemd/blkid does not help us to retrieve all needed information and seems that writing own parsing (either via /sys or via direct access to disk device) seems to be the only option which would cover all required bits.

Anyway, currently for me it looks like that it is not something which could block merging this PR.

fstirlitz:
> Ok, but systemd nor blkid does not help us with detection of such setup. For hybrid MBR setup you need to parse both MBR and GPT sectors manually.

I don't think it needs to be handled _that_ precisely. I think it's safe to assume that if a partition is defined by a GPT disklabel, then any hypothetical hybrid MBR will treat it as a primary partition (because it'd be hard to squeeze in an EBR in such a setup), so no adjustment is necessary for that case. It should be enough to check for extended partitions in a pure MBR setup (with appropriate partition IDs), and treat all other cases the same way as a primary MBR partition, i.e. fill the field with the absolute disk offset of the partition.

> But currently it looks like that systemd/blkid does not help us to retrieve all needed information and seems that writing own parsing (either via /sys or via direct access to disk device) seems to be the only option which would cover all required bits.

You'd still have to discover the VFS node of the parent device of the partition, and udev is the natural place to ask for it; it will give you the right answer even for symbolic links and device nodes manually created with `mknod`.

> are you going to fully finish your investigation how it works?

Probably yes. I'm planning to create a handful of variously partitioned disk images, format them under various versions of DOS and compare the results. I'm not sure yet when I find the time for it, but I expect it should be sooner rather than later. I will then report my findings in #29.

pali:
> I don't think it needs to be handled _that_ precisely. I think it's safe to assume that if a partition is defined by a GPT disklabel, then any hypothetical hybrid MBR will treat it as a primary partition (because it'd be hard to squeeze in an EBR in such a setup)

IIRC, gdisk can create such hybrid MBR.

> > But currently it looks like that systemd/blkid does not help us to retrieve all needed information and seems that writing own parsing (either via /sys or via direct access to disk device) seems to be the only option which would cover all required bits.
> 
> You'd still have to discover the VFS node of the parent device of the partition

Retrieving parent device with this PR should be easy.

> I will then report my findings in #29.

Ok.

pali:
After rebasing, I'm merging it in current state. In future there can be other improvements.

PR #106, opened 2018-12-05
Use SOURCE_DATE_EPOCH for reproducible builds

bjornfor:
Hi,

Here are some patches that implement reproducible builds with the SOURCE_DATE_EPOCH variable (in the context of dosfstools being used as part of builds).

I've run the test suite and it passes.

### DIFF
--- a/src/boot.c
+++ b/src/boot.c
@@ -33,6 +33,8 @@
 #include <stdlib.h>
 #include <sys/types.h>
 #include <time.h>
+#include <errno.h>
+#include <ctype.h>
 
 #include "common.h"
 #include "fsck.fat.h"
@@ -672,6 +674,7 @@ void write_volume_label(DOS_FS * fs, char *label)
 {
     time_t now;
     struct tm *mtime;
+    char *source_date_epoch = NULL;
     off_t offset;
     int created;
     DIR_ENT de;
@@ -687,8 +690,24 @@ void write_volume_label(DOS_FS * fs, char *label)
     if (de.name[0] == 0xe5)
 	de.name[0] = 0x05;
 
-    now = time(NULL);
-    mtime = (now != (time_t)-1) ? localtime(&now) : NULL;
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *tmp = NULL;
+        long long conversion = 0;
+        errno = 0;
+        conversion = strtoll(source_date_epoch, &tmp, 10);
+        now = conversion;
+        if (!isdigit((unsigned char)*source_date_epoch) || *tmp != '\0'
+                || errno != 0 || (long long)now != conversion) {
+            die("SOURCE_DATE_EPOCH is too big or contains non-digits: \"%s\"",
+                source_date_epoch);
+        }
+        mtime = gmtime(&now);
+    } else {
+        now = time(NULL);
+        mtime = (now != (time_t)-1) ? localtime(&now) : NULL;
+    }
+
     if (mtime && mtime->tm_year >= 80 && mtime->tm_year <= 207) {
 	de.time = htole16((unsigned short)((mtime->tm_sec >> 1) +
 					   (mtime->tm_min << 5) +
--- a/src/common.c
+++ b/src/common.c
@@ -30,6 +30,7 @@
 #include <string.h>
 #include <stdarg.h>
 #include <errno.h>
+#include <ctype.h>
 #include <wctype.h>
 #include <termios.h>
 #include <sys/time.h>
@@ -298,8 +299,21 @@ void check_atari(void)
 uint32_t generate_volume_id(void)
 {
     struct timeval now;
+    char *source_date_epoch = NULL;
 
-    if (gettimeofday(&now, NULL) != 0 || now.tv_sec == (time_t)-1 || now.tv_sec < 0) {
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *tmp = NULL;
+        long long conversion = 0;
+        errno = 0;
+        conversion = strtoll(source_date_epoch, &tmp, 10);
+        if (!isdigit((unsigned char)*source_date_epoch) || *tmp != '\0'
+                || errno != 0) {
+            die("SOURCE_DATE_EPOCH is too big or contains non-digits: \"%s\"",
+                source_date_epoch);
+        }
+        return (uint32_t)conversion;
+    } else if (gettimeofday(&now, NULL) != 0 || now.tv_sec == (time_t)-1 || now.tv_sec < 0) {
         srand(getpid());
         /* rand() returns int from [0,RAND_MAX], therefore only 31 bits */
         return (((uint32_t)(rand() & 0xFFFF)) << 16) | ((uint32_t)(rand() & 0xFFFF));
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -1455,7 +1455,6 @@ static void usage(const char *name, int exitval)
     fprintf(stderr, "  -v              Verbose execution\n");
     fprintf(stderr, "  --variant=TYPE  Select variant TYPE of filesystem (standard or Atari)\n");
     fprintf(stderr, "\n");
-    fprintf(stderr, "  --invariant     Use constants for randomly generated or time based values\n");
     fprintf(stderr, "  --offset=SECTOR Write the filesystem at a specific sector into the device file.\n");
     fprintf(stderr, "  --help          Show this help message and exit\n");
     exit(exitval);
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -1074,7 +1074,7 @@ static void setup_tables(void)
         }
 
         /* If is not available then generate random 32 bit disk signature */
-        if (invariant)
+        if (invariant || getenv("SOURCE_DATE_EPOCH"))
             disk_sig = volume_id;
         else if (!disk_sig)
             disk_sig = generate_volume_id();
@@ -1287,7 +1287,7 @@ static void setup_tables(void)
 	    de->name[0] = 0x05;
 	de->attr = ATTR_VOLUME;
 	if (create_time != (time_t)-1) {
-	    if (!invariant)
+	    if (!invariant && !getenv("SOURCE_DATE_EPOCH"))
 		ctime = localtime(&create_time);
 	    else
 		ctime = gmtime(&create_time);
@@ -1477,6 +1477,7 @@ int main(int argc, char **argv)
     int blocks_specified = 0;
     struct timeval create_timeval;
     long long conversion;
+    char *source_date_epoch = NULL;
 
     enum {OPT_HELP=1000, OPT_INVARIANT, OPT_MBR, OPT_VARIANT, OPT_CODEPAGE, OPT_OFFSET};
     const struct option long_options[] = {
@@ -1497,8 +1498,20 @@ int main(int argc, char **argv)
 	    program_name = p + 1;
     }
 
-    if (gettimeofday(&create_timeval, NULL) == 0 && create_timeval.tv_sec != (time_t)-1)
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        errno = 0;
+        conversion = strtoll(source_date_epoch, &tmp, 10);
+        create_time = conversion;
+        if (!isdigit((unsigned char)*source_date_epoch) || *tmp != '\0'
+                || errno != 0 || (long long)create_time != conversion) {
+            die("SOURCE_DATE_EPOCH is too big or contains non-digits: \"%s\"",
+                source_date_epoch);
+        }
+    } else if (gettimeofday(&create_timeval, NULL) == 0 && create_timeval.tv_sec != (time_t)-1) {
         create_time = create_timeval.tv_sec;
+    }
+
     volume_id = generate_volume_id();
     check_atari();
 
--- a/tests/test-mkfs
+++ b/tests/test-mkfs
@@ -52,7 +52,7 @@ echo "Test $testname"
 rm -f "${testname}.out" "${testname}.refimg"
 
 xxd -r "${srcdir}/${testname}.xxd" "${testname}.refimg" || exit 99
-run_mkfs -C -v --invariant $ARGS "${testname}.out" $SIZE || exit 99
+(export SOURCE_DATE_EPOCH=1426325213; run_mkfs -C -v -i 0x1234abcd $ARGS "${testname}.out" $SIZE) || exit 99
 
 echo
 echo "Comparing..."


bjornfor:
Hm, the Travis CI build fails during "make check", while it passes on my machine. I guess I'll have to experiment with travis by doing a number of force-pushes.

andreasbombe:
Admittedly, when I created the `--invariant` flag I didn't think of using the reproducible build variables for that. I'm not sure how I feel about removing that flag, yet.

Though the tests are failing on the reference filesystem tests, so something there isn't setting the time quite correctly.

bjornfor:
I figured it out. "SOURCE_DATE_EPOCH=1426325213 command" doesn't work in dash, only bash. I changed it to "export SOURCE_DATE_EPOCH=1426325213; command" which works in both shells.

@andreasbombe: If you don't like deprecating (and eventually removing) the --invariant flag, then you can pick only the first commit. Or I can remove them and force-push, if you like.

src/boot.c
```diff
@@ -690,10 +690,16 @@ static void write_volume_label(DOS_FS * fs, char *label)
 {
     time_t now = time(NULL);
     struct tm *mtime = localtime(&now);
+    char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
     off_t offset;
     int created;
     DIR_ENT de;
 
+    if (source_date_epoch) {
+	    now = strtoull(source_date_epoch, NULL, 10);
```
pali: Does not check for garbage nor correct data in string. Also usage of strtoull is dangerous as it does not signal underflow. strtoll should be used. See also: https://github.com/dosfstools/dosfstools/pull/92

src/mkfs.fat.c
```diff
@@ -1170,9 +1170,14 @@ static void setup_tables(void)
 	if (de->name[0] == 0xe5)
 	    de->name[0] = 0x05;
 	de->attr = ATTR_VOLUME;
-	if (!invariant)
-		ctime = localtime(&create_time);
-	else
+	if (!invariant) {
+		char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+		if (source_date_epoch) {
+			create_time = strtoull(source_date_epoch, NULL, 10);
```
pali: Same there.

src/boot.c
```diff
@@ -690,10 +692,29 @@ static void write_volume_label(DOS_FS * fs, char *label)
...
     struct tm *mtime = localtime(&now);
+    char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
     off_t offset;
     int created;
     DIR_ENT de;
 
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        errno = 0;
+        now = strtoll(source_date_epoch, &endptr, 10);
```
pali: `now` is of type `time_t` which is 32bit on 32bit systems. `strtoll` returns `long long` value (at least 64bit) which overflow when storing into 32bit type.

src/check.c
```diff
@@ -143,6 +143,10 @@ static char *file_stat(DOS_FILE * file)
     date =
 	date_dos2unix(le16toh(file->dir_ent.time), le16toh(file->dir_ent.date));
     tm = localtime(&date);
+    if (!tm) {
```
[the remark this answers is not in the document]
bjornfor: Ok, I'll remove that commit.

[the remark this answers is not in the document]
bjornfor: Oh, it's not just that extra commit, its the hole point of the PR (SOURCE_DATE_EPOCH) that conflicts with #94.

src/boot.c
bjornfor: Writing this correctly is difficult. There is no TIME_T_MAX to check against, and apparently time_t can be any integer or float type. And even as integer type, it can have "padding bits", so one cannot check the max possible value based on sizeof(time_t).

However, if we assume time_t is long int, which seems to be the most common type, how about this?

```
time_t now = 0;
long long val = strtoll(...);
/* check for strtoll success */

if (sizeof(now) == sizeof(int32_t)) {
    if (val > INT32_MAX) {
        /* overflow, abort */
    } else {
        now = val;
    }
} else if (sizeof(now) >= sizeof(val)) {
    now = val;
} else {
    /* Unsupported time_t type/size, abort. */
    /* Or add build time assert about time_t type/size? */
}
```

bjornfor:
Reminder to self: there is a gettimeofday() call in src/common.c that I must patch too.

src/boot.c
pali: In your example you have a problem with signedness, when now and val differs.

In past I proposed following check for underflow/overflow:
```c
long long val;
...
time_t now = val;
if (((time_t)-1 > 0 && val < 0) || now > LLONG_MAX || (long long)now != val) {
    /* conversion failed */
}
```

If you want to be pedant and check also floating point time_t type, then you can use:
```c
if ((time_t)1 / 2 == 0) {
    /* time_t is integer */
} else {
    /* time_t is floating point
}
```

In case it is floating point, you can probably ignore problems with casting as in most cases you loose precision. Or use one of `strtof`/`strtod`/`strtold` function.

bjornfor: Nice and compact. I like it! But I don't understand how ```now > LLONG_MAX``` can ever be true, after the ```time_t now = val;``` line. Care to explain?

pali: This is theoretical case, when time_t is 64bit unsigned type.

Basically it is safe check prior using `(long long)now`, to check that casting does not overflow. I'm doing these checks prior comparing and casting in any code.

bjornfor: When building for 32-bit system, gcc prints this:

```
warning: comparison is always false due to limited range of data type [-Wtype-limits]
             if (((time_t)-1 > 0 && val < 0) || now > LLONG_MAX || (long long)now != val) {
                                                    ^
```

What's a nice way to deal with that warning?

pali: It is wrong/false-positive warning. Reason why warnings in production builds are turned off...
You can compile with -Wno-type-limits or add some gcc pragma which turns that warning locally (I do not remember how to control them right now).

pali: Or you can try to play with `__builtin_choose_expr` or with `__builtin_constant_p`. See:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fchoose_005fexpr
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fconstant_005fp

pali: Or drop that `now > LLONG_MAX` check. Negative `val` with unsigned `time_t` is already checked by previous condition.

bjornfor: Ah, even better, no platform specific code.

Since I'm checking that the first character starts with a digit, I know the result can not be negative. I think this will suffice then:

```
long long val = strtoll(source_date_epoch, &endptr, 10);
time_t create_time = val;
if (!isdigit(*source_date_epoch) || *endptr != '\0' || errno != 0 || (long long)create_time != val) {
    /* conversion failed */
}
```

(I'm not handling time_t as float - I'm not that pedantic.)

bjornfor:
@pali: Rebased. I did the conversion validation as mentioned in https://github.com/dosfstools/dosfstools/pull/106#discussion_r241975324. Please have a look.

src/boot.c
```diff
@@ -705,8 +708,24 @@ static void write_volume_label(DOS_FS * fs, char *label)
     if (de.name[0] == 0xe5)
 	de.name[0] = 0x05;
 
-    now = time(NULL);
-    mtime = (now != (time_t)-1) ? localtime(&now) : NULL;
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        errno = 0;
+        long long val = strtoll(source_date_epoch, &endptr, 10);
```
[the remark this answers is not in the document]
bjornfor: Oh, I saw both styles so I just picked this one. I'll change it.

src/common.c
```diff
@@ -296,8 +297,22 @@ void check_atari(void)
...
-
-    if (gettimeofday(&now, NULL) != 0 || now.tv_sec == (time_t)-1 || now.tv_sec < 0) {
+    char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        errno = 0;
+        long long val = strtoll(source_date_epoch, &endptr, 10);
+        now.tv_sec = val;
+        now.tv_usec = 0;
```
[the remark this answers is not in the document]
bjornfor: Ah, I missed that. Will set ```now.tv_usec = val``` on next push.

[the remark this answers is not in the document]
pali: Why `tv_usec` to `val`? `val` value for seconds, not for microseconds.

You can simplify this problem to just:
```c
if (source_date_epoch) {
    char *endptr = NULL;
    errno = 0;
    long long val = strtoll(source_date_epoch, &endptr, 10);
    if (!isdigit(*source_date_epoch) ||  *endptr != '\0' || errno != 0) {
        /* error */
    }
    return (uint32_t)val;
}
```

src/boot.c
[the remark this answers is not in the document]
bjornfor: Declare before use, but in smallest possible scope, or at function level? Initialize variables at definition time?

src/common.c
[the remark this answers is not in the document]
bjornfor: > Why tv_usec to val? val value for seconds, not for microseconds.

I thought the actual value didn't matter. I'll change it.

tests/test-mkfs
```diff
@@ -52,7 +52,7 @@ echo "Test $testname"
 rm -f "${testname}.out" "${testname}.refimg"
 
 xxd -r "${srcdir}/${testname}.xxd" "${testname}.refimg" || exit 99
-run_mkfs -C -v --invariant $ARGS "${testname}.out" $SIZE || exit 99
+(export SOURCE_DATE_EPOCH=1426325213; run_mkfs -C -v -i 0x1234abcd $ARGS "${testname}.out" $SIZE) || exit 99
```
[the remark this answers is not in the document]
bjornfor: The tests break if using anything but the specific timestamp and volume_id, as given by the ```--invariant``` flag they replace. See here: https://github.com/dosfstools/dosfstools/blob/a8746502cb28ee28a1bbbbcef089b35566103bdc/src/mkfs.fat.c#L1592-L1595.

I didn't investigate further, I simply continued the existing behaviour.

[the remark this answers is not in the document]
pali: So `invariant` now use volume id 0x1234abcd. So update volume id in testing image to match volume id computed from SOURCE_DATE_EPOCH? This could be more cleaner way.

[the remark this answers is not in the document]
bjornfor: I actually like it the way it is; the volume-id is specified as a hex number, so it's easier to spot in the reference image (.xxd file). (SOURCE_DATE_EPOCH is specified in base 10.) And the separate flag provides a separate knob to use in testing (although it is hardcoded in the tests right now.)

Also, I'm not sure @andreasbombe would even like to merge that commit, given [this comment](https://github.com/dosfstools/dosfstools/pull/106#issuecomment-444705187). I'm sort of waiting for confirmation whether I should remove that commit.

I can update the reference images and remove the use of volume-id flag in the test, if that is preferred, but I just thought I'd say what's on my mind first.

bjornfor:
I pushed another update:

    Changes v3 -> v4:
    * Use die() for fatal errors, instead of fprintf() + exit().
    * Fix bug from v3 where volume-id was calculated with 20 bits of zero
      instead of real value bits from SOURCE_DATE_EPOCH.
    * Refactoring / review feedback.

Please have a look.

src/common.c
```diff
@@ -296,8 +297,20 @@ void check_atari(void)
...
+    char *source_date_epoch = NULL;
+
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        long long val = 0;
+        errno = 0;
+        val = strtoll(source_date_epoch, &endptr, 10);
+        if (!isdigit(*source_date_epoch) || *endptr != '\0' || errno != 0) {
+            die("SOURCE_DATE_EPOCH is too big or contains non-digits: \"%s\"\n",
```
pali: Same there

src/mkfs.fat.c
```diff
@@ -1371,8 +1372,22 @@ int main(int argc, char **argv)
...
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        long long val = 0;
+        errno = 0;
+        val = strtoll(source_date_epoch, &endptr, 10);
+        create_time = val;
+        if (!isdigit(*source_date_epoch) || *endptr != '\0' || errno != 0
+                || (long long)create_time != val) {
+            die("SOURCE_DATE_EPOCH is too big or contains non-digits: \"%s\"\n",
```
pali: And same there

bjornfor:
@pali: I pushed a new rebase where trailing '\n' in calls to die() have been removed.

bjornfor:
Ping.

tests/test-mkfs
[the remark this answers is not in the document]
pali: But this does not test that `SOURCE_DATE_EPOCH` env generates stable volume id.

pali:
Also remove changelog from commit message as it is not needed there.

tests/test-mkfs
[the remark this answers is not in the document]
bjornfor: What did you have in mind? To run the test one more time, with different SOURCE_DATE_EPOCH and reference image? I didn't do that because I just wanted to update the test from the special --invariant flag to the generic SOURCE_DATE_EPOCH. I didn't want to make functional changes to the test.

Also, @andreasbombe hasn't replied whether they would like to even deprecate the --invariant flag. So the two last commits might get dropped.

bjornfor:
> Also remove changelog from commit message as it is not needed there.

Right. That was getting excessive. Fixed.

tests/test-mkfs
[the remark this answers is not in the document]
pali: > What did you have in mind?

I mean that you are using `SOURCE_DATE_EPOCH` in test, but then volume id is not generated by `SOURCE_DATE_EPOCH`, but overridden by `-i` option.

So it does not test that `SOURCE_DATE_EPOCH` is fully working. What I mean is to use only `SOURCE_DATE_EPOCH` as I wrote in first post.

@andreasbombe has not replied yet, but personally I do not see reason why to not use `SOURCE_DATE_EPOCH` in test if you are introducing it. And I guess that `--invariant` option is there specially only for tests as I do not see any benefit to have static volume id. Static hardcoded volume id just cause problems as it is used as unique identifier.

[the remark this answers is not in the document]
bjornfor: > I mean that you are using `SOURCE_DATE_EPOCH` in test, but then volume id is not generated by `SOURCE_DATE_EPOCH`, but overridden by `-i` option.

Right, volume_id is overridden. But not create_time.

If I remove the -i flag then I have to update the reference image to make the test pass. That means:
* There is already (some) test coverage for SOURCE_DATE_EPOCH.
* Replacing -i with SOURCE_DATE_EPOCH (+ updating ref. image) would mean removing test coverage for the -i code path.

I guess instead of updating the existing tests, we can add new tests with SOURCE_DATE_EPOCH=X and ref. image X, where X=[1, 2]? But personally I feel the coverage is OK as is.

bjornfor:
Rebased to resolve conflict after recent changes to master.

src/mkfs.fat.c
```diff
@@ -1372,8 +1373,22 @@ int main(int argc, char **argv)
 	    program_name = p + 1;
     }
 
-    if (gettimeofday(&create_timeval, NULL) == 0 && create_timeval.tv_sec != (time_t)-1)
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *endptr = NULL;
+        long long val = 0;
+        errno = 0;
+        val = strtoll(source_date_epoch, &endptr, 10);
```
pali: Use `conversion` and `tmp` variables like other `strtoll` usage in this file.

bjornfor: Fixed.

src/mkfs.fat.c
```diff
@@ -1372,8 +1373,22 @@ int main(int argc, char **argv)
 	    program_name = p + 1;
     }
 
-    if (gettimeofday(&create_timeval, NULL) == 0 && create_timeval.tv_sec != (time_t)-1)
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *tmp = NULL;
+        long long conversion = 0;
```
pali: You do not need to declare `tmp` and `conversion` again. It is on line 1347 and 1356.

bjornfor: Oh, right! Fixed.

tests/test-mkfs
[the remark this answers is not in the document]
pali: We do not have answer from @andreasbombe about removal of `--invariant` option, but I'm for deprecation in next release and later to fully remove it. Therefore I'm fine with removing it from documentation. But also it should be removed from all existing usage, therefore from tests too.

`SOURCE_DATE_EPOCH` seems like a standard solution how to deal with stable timestamps and `--invariant` option is just dosfstools specific.

So I'm for:
* removing `--invariant` option from help and documentation
* updating test scripts and test files to use `SOURCE_DATE_EPOCH`
* do not use `--invariant` in tests anymore
* let `--invariant` to be work in next release with release notes information about deprecation
* removal of `--invariant` option from code after next release

bjornfor:
Rebased on master to resolve new conflicts.

bjornfor:
Hi, is there anything I can do to get this merged?

pali:
@bjornfor I'm fine with first two commits. And about deprecation we have not get answer from @andreasbombe (my proposal is in https://github.com/dosfstools/dosfstools/pull/106#discussion_r259586566)

pali:
@bjornfor: As there is no answer from @andreasbombe, I'm going to accept all 3 patches into master branch. Can you rebase this pull request on top of the master branch?

bjornfor:
@pali: Thanks! Rebased on master now (no conflicts).

pali:
@bjornfor: tests are not passing :-( can you look at it?

pali:
@bjornfor: Are you going to look at these issues? (Also more other pull requests were merged into master, therefore a new rebasing would be required)

bjornfor:
I had a quick look, was not able to fix it (or understand why it broke) and I haven't had time to look more into it.

pali:
Ok, I'm letting this pull request open.

pali:
@bjornfor or anybody else: Are you going to fix this pull request?

hieunv-tsdv:
@bjornfor @pali ,

I checked this pull request, it seems the problem is coming from the "random disk signature".
The mkfs.fat should not generate a random disk signature when "SOURCE_DATE_EPOCH" and volume_id already exists, similar to using "--invariant" option.
You can see all the CI are PASS after I fixed it:
> https://github.com/hieunv-tsdv/dosfstools/commit/fbb5e91e4fec8fb73ed6e9d9099f03c9b1f79525
> https://travis-ci.com/github/hieunv-tsdv/dosfstools

bjornfor:
@hieunv-tsdv: Thank you! I've squashed your fix into my originally broken commit and rebased onto master. Now the tests pass!

pali:
Hello! Tests were not started on github/travis. Could you please rebase commits on top of master branch? After new push tests should be started again automatically.

bjornfor:
Rebased.

src/mkfs.fat.c
```diff
@@ -1074,7 +1074,7 @@ static void setup_tables(void)
         }
 
         /* If is not available then generate random 32 bit disk signature */
-        if (invariant)
+        if (invariant || (getenv("SOURCE_DATE_EPOCH") && volume_id))
             disk_sig = volume_id;
         else if (!disk_sig)
             disk_sig = generate_volume_id();
```
pali, line 1080: I'm looking at this change and I would like to know: Why check for `volume_id` is needed here? On line 514 is: `volume_id = generate_volume_id();` And `generate_volume_id()` introduced in this patch returns value from `getenv("SOURCE_DATE_EPOCH")` if this env is set.

hieunv-tsdv: Hi @pali ,

> And generate_volume_id() introduced in this patch returns value from getenv("SOURCE_DATE_EPOCH") if this env is set

Yes. 
However, the _volume_id_ can be passed from "-i" option. I'm not sure if an user passed in "-i" option a null value.
We can remove the check for _volume_id_ if we make sure that "-i" cannot pass a null value.

```c
case 'i':               /* i : specify volume ID */
            errno = 0;
            conversion = strtoll(optarg, &tmp, 16);
        
            if (!*optarg || isspace(*optarg) || *tmp || conversion < 0) {
                printf("Volume ID must be a hexadecimal number\n");
```

pali: Well, I have not find any restriction to volume id number. So I think that **zero** is also valid volume id.

Microsoft fatgen103 specification says about volume id:
> This ID is usually generated by simply combining the current date and time into a 32-bit value.

So I think that zero volume id should be allowed.

Also FAT12 and FAT16 allows to indicate that volume id is not recorded at all and therefore allow to distinguish between _volume id is set to 0_ and _volume id is not set_.

hieunv-tsdv:
@pali: Thanks for the information.

@bjornfor: Could you please help me update this merge request:
`
if (invariant || (getenv("SOURCE_DATE_EPOCH") && volume_id))
=> if (invariant || getenv("SOURCE_DATE_EPOCH"))
`
This change doesn't affect to the CI jobs: https://travis-ci.com/github/hieunv-tsdv/dosfstools/builds/235387092
sorry for this inconvenience.

bjornfor:
@hieunv-tsdv @pali: Updated diff.

src/boot.c
```diff
@@ -687,8 +690,24 @@ void write_volume_label(DOS_FS * fs, char *label)
...
-    now = time(NULL);
-    mtime = (now != (time_t)-1) ? localtime(&now) : NULL;
+    source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch) {
+        char *tmp = NULL;
+        long long conversion = 0;
+        errno = 0;
+        conversion = strtoll(source_date_epoch, &tmp, 10);
+        now = conversion;
+        if (!isdigit(*source_date_epoch) || *tmp != '\0' || errno != 0
```
pali, line 700: And now I see there another issue. `*source_date_epoch` is of type `char`, which means `signed char` on x86, so contains values from range `[-128, 127]` (values `[128, 255]` are encoded as negative). But function `isdigit()` takes type `int` and expects characters to be in range `[0, 255]`, so this function is incompatible with `signed char` type.

So explicit cast to `unsigned char` is required prior passing value into `isdigit()`. This trap is documented also in manpage, in NOTES section: https://man7.org/linux/man-pages/man3/isdigit.3.html

```suggestion
        if (!isdigit((unsigned char)*source_date_epoch) || *tmp != '\0' || errno != 0
```

bjornfor: @pali: Thanks! I updated the 3 "isdigit()" calls with that cast.

pali:
Thanks! Merged.

src/boot.c
pali: I fixed this trap also for other is* functions.

PR #113, opened 2019-01-14
Fixed off-by-one in local_string_to_dos_string.

stoeckmann:
The function local_string_to_dos_string is vulnerable to an off-by-one
buffer overflow. In fact, it is triggered in default usage and becomes
visible when compiled with ASAN:

$ CFLAGS="-fsanitize=address" ./configure
$ dd if=/dev/zero of=example.iso bs=1024 seek=64 count=1
$ ./src/mkfs.fat example.iso
mkfs.fat 4.1+git (2017-01-24)

=================================================================
==3857==ERROR: AddressSanitizer: stack-buffer-overflow on address ...

The problem is that the argument "len" to local_string_to_dos_string
stores the length of the output buffer. Yet it can also be used as an
index to store '\0':

If the whole "out" buffer has been written to, bytes_out is 0 and
the assignment in out[len-bytes_out] therefore leads to an off-by-one.

Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>

### DIFF
--- a/src/charconv.c
+++ b/src/charconv.c
@@ -87,6 +87,6 @@ int local_string_to_dos_string(char *out, char *in, unsigned int len)
                 in, used_codepage);
         return 0;
     }
-    out[len-bytes_out] = 0;
+    out[len-1-bytes_out] = 0;
     return 1;
 }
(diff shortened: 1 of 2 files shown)


pali:
Nice catch! Can you rename `len` variable to `out_size`? So variable name better match its purpose, it is not length of string (which is usually without nul term), but rather buffer size (which is usually including nul term).

stoeckmann:
Makes sense. If it's not wanted, the first commit can be cherry-picked.

PR #114, opened 2019-01-17
Fix out of boundary read in fsck.fat/fatlabel

stoeckmann:
Faulty filesystems are able to trigger integer overflows in read_boot which eventually lead to insufficient allocation of memory for the FAT and therefore out of boundary reads.

How to trigger this OOB read (and eventually a segmentation fault):

$ dd if=/dev/zero of=fat32.iso bs=1M seek=100 count=1
$ mkfs.fat -f 1 -F 32 -S 512 fat32.iso
$ base64 -d <<< IQAACgAAAAI= | dd of=fat32.iso bs=1 seek=32 conv=notrunc
$ fsck.fat fat32.iso

I have kept the FAT backup intact, so you have to choose "3" to continue without modifications. It's easiest to compare both FATs this way.

Took a while to trigger an out of boundary read on an i686 machine so you might want to compile with ASAN as in CFLAGS="-fsanitize=address" ./configure.


pali:
Hi! In dosfstools in most cases for 64bit arithmetic is used `long long` type (or `unsigned long long`). Could you follow this coding style?

src/boot.c
```diff
@@ -451,9 +452,10 @@ void read_boot(DOS_FS * fs)
     struct boot_sector b;
     unsigned total_sectors;
     unsigned int logical_sector_size, sectors;
-    off_t fat_length;
+    long long fat_length;
     unsigned total_fat_entries;
     off_t data_size;
+    long long position;
```
pali: It should be unsigned as you are doing assignment from unsigned values.

src/boot.c
```diff
@@ -478,23 +480,34 @@ void read_boot(DOS_FS * fs)
     if (verbose)
 	printf("Checking we can access the last sector of the filesystem\n");
     /* Can't access last odd sector anyway, so round down */
-    fs_test((off_t)((total_sectors & ~1) - 1) * logical_sector_size,
-	    logical_sector_size);
+    position = (long long)((total_sectors & ~1) - 1) * logical_sector_size;
+    if (position != (off_t)position)
```
pali: I'm thinking about defining `OFF_MAX` value as:
```c
#include <limits.h>
#ifndef OFF_MAX
#define OFF_MAX ((1ULL << (sizeof(off_t) * CHAR_BIT - 1)) - 1)
#endif
```
And then comparing `position` against it. But I'm not sure if it is a good idea...

stoeckmann:
It is an assignment of a signed type, yes. But it's been like that all the time due to off_t being signed as well.

src/boot.c
stoeckmann: In fact, the assignments are already casted to long long explicitly or implicitly. I see no concern here because the data type is large enough to hold all possible values.

pali: You are doing assignment from `((total_sectors & ~1) - 1) * logical_sector_size` and this is always unsigned. So I do not think there is need to store non-negative value into signed type if there is no need to represent negative values.

stoeckmann: So you prefer to have an unsigned position which is used to assign off_t's later on? I don't get the point here.

stoeckmann: The result of `((total_sectors & ~1) - 1)` is explicitly casted to `(long long)` which is save as long as unsigned is smaller than long long in terms of available bits. The multiplication with logical_sector_size is therefore signed and long long as well.

pali: Hm... I'm thinking about it... `total_sectors` is unsigned and maximally 32bit. `logical_sector_size` is unsigned and maximally 16bit. Result is unsigned, so should be stored in `unsigned long long`. But also it always fit into 24bits, therefore also `signed long long` type is suitable (after casting). But you are right that `off_t` is signed and in most cases up to the `2^63-1` or `2^31-1`.

Personally I prefer to not use "less" ambiguous types (like `off_t` or `time_t`) and rather `unsigned long long` which is guaranteed to be at least 64 bit long.

And in case that value must not be negative (like number of sectors, sector size or offset from beginning of file) then I rather prefer to use unsigned type to enforce this constraint. Usage of signed type tells reader "this is signed type, it can be negative, so be careful".

stoeckmann: I agree here and I would have used uint64_t for that (and later on switched to unsigned long long) because negative values wouldn't make any sense for these values, after all they are unsigned in FAT specification as well.

But due to usage of off_t throughout the code and especially for storing these values in structs I've kept using a signed type. In fact, on linux/amd64 these changes are basically no-ops because off_t == long long.

This keeps my changes as small as possible in terms of "change of behaviour" on amd64 and increases the width of data types far enough on these few systems with off_t being 32 bit, i.e. linux/i686 and others with --disable-largefile.

pali: Probably the best fix would be to use unsigned values in structs and convert them to `off_t` just before passing them into `lseek` function.

So we can let decision to @andreasbombe how he wants to have this problem handled/fixed.

Current usage of types is really mess and I understand that you want ot keep changes as small as possible.

pali:
In pull request https://github.com/dosfstools/dosfstools/pull/115 I added `-fsanitize=address` into travis build check and under clang it really failed. This can help extending test suite.

src/boot.c
stoeckmann: Yes, that sounds great. In that case read_boot can fully validate that the boot sector is correct even on systems that couldn't handle the file system later on due to size limitations.

And such a change should definitely go into its own merge request...

src/boot.c
pali: @stoeckmann do you have any opinion if above is a good idea or not?

pali:
Pull request https://github.com/dosfstools/dosfstools/pull/115 which enable automatic testing on Travis with `-fsanitize=address` was merged and there are no current problems. Could you extend test coverage  also with your discovered out of boundary read? Seems that dosfstools would need better test coverage.

src/boot.c
stoeckmann: Sorry for not getting back at this in a timely manner. I've already tested it just to see if it works correctly and it does.

I'm fine with that approach. It's just that I tried to avoid implementing it like that for assuming that there is a good reason that OFF_MAX is not part of limits.h.

pali: Looks like that some systems provides `OFF_MAX` macro in limits.h (e.g. freebsd). Some software defines own `OFF_T_MAX` macro.

stoeckmann: Cool, that makes it even more obvious to be a good solution. I'll adjust my pull request.

pali: Anyway, in pull request https://github.com/dosfstools/dosfstools/pull/84#pullrequestreview-202281703 is exactly same problem, checking if value in long long can be stored into off_t.

So we should choose one solution for this `off_t` problem and use it on both places.

So maybe `OFF_MAX` should be defined in some header file?

stoeckmann:
I've added a test, but XFAIL doesn't mean much. It is supposed to return with an error exit code, but even if it crashes, it will do.

Is there a sane way to add an expected error code? Or at least say that values above 128 are not okay?

src/boot.c
stoeckmann: In that case it should be done in an own refactoring commit. At first commons.h looked like a good spot, but that file itself doesn't even contain off_t. Adding off_t to fsck.fat.h would have made sense, because that file uses off_t in the offending fashion. But then again it would have to be repeated at least twice as well.

After seeing the definition of TRUE and FALSE in mkfs.fat.c I decided to not play around with OFF_MAX much further, because these adjustments better fit into own commits.

pali: `src/common.h` is included in all source files, so seems as the best place for defining OFF_MAX. It is not a problem that file does not have any function with `off_t` type. So please put it there. Defining OFF_MAX on two places does make sense.

TRUE/FALSE macros are just some historic code and candidate for future removal...

pali:
Why you marked test as XFAIL?

stoeckmann:
The patch will still return an error value because the file system is too huge (but legal) to be processed with current implementation.

Therefore it makes sense that fsck.fat fails (without segfault) to indicate that the file system cannot be processed.

pali:
> The patch will still return an error value because the file system is too huge (but legal) to be processed with current implementation.

So fsck.fat has problems to process too huge FAT filesystems which are according to specification?

stoeckmann:
FAT32 could be 2 TB in size, which definitely cannot be addressed with an unsigned int (32 bit). So yes, this is currently a reason that fsck.fat cannot handle huge FAT32 filesystems.

pali:
FAT32 can be also larger then 2TB when cluser size is bigger (e.g. on Native 4K disks). So this is definitely a bug. Would you like to fix it? :-) Or at least report an issue with reproducer?

src/boot.c
stoeckmann: Added to src/common.h as requested.

stoeckmann:
I'll add a bug report with references to these OFF_MAX occurrences in boot.c after this is merged. Theses occurrences already indicate what has to be fixed.

src/common.h
```diff
@@ -23,8 +23,13 @@
 #ifndef _COMMON_H
 #define _COMMON_H
 
+#include <limits.h>
 #include <stdint.h>
 
+#ifndef OFF_MAX
+	#define OFF_MAX ((1ULL << (sizeof(off_t) * CHAR_BIT - 1)) - 1)
```
pali: You need to `#include <sys/types.h>` as `off_t` is defined in this file.

Also I think that result of above `((1ULL ...` should be casted to `off_t` as now `1ULL` results in `unsigned` type (`off_t` is signed) which may produce false-positive warnings when comparing with some other variable of `off_t` type.

stoeckmann: I would conclude that this should be properly prepared and not be part of this merge then.

I'll keep this merge request as it is, people are free to cherry pick what's wanted and leave out whatever is still worth to be discussed.

pali: It is just a simple fix. If you agree I could try to use github feature to edit file in web browser.

stoeckmann: Done.

pali:
Ok, fine, test passed!

Just squash relevant commits (e.g. in one you introduced new int64_t variables and in later you changed it to long long; or started using OFF_MAX) and I merge this pull request.

stoeckmann:
As all of these commits are needed to fix handling of large file systems, all of them can be squashed. So doing it for the whole merge request is the easiest way.

pali:
Ok, I have there a github button "Squash and merge". We will see if github is working...

stoeckmann:
Yes, it worked. :)

pali:
great!

PR #115, opened 2019-01-19
Update Travis configuration

pali:
* Use both gcc and clang compilers
* Compile with -fwrapv and -fsanitize=address
* Compile in both 32 and 64 modes for x86 systems
* Cross compile for little endian arm and big endian powerpc
* Run cross compiled binaries in qemu

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,66 @@
 language: c
 
-before_script: autoreconf -i
-script: ./configure && make all check
+compiler:
+  - gcc
+  - clang
+
+env:
+  - CFLAGS="-W -Wall"
+  - CFLAGS="-W -Wall -O2 -g -fwrapv -fsanitize=address"
+  - CFLAGS="-m32 -W -Wall"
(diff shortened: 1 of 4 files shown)


pali:
Seems that Travis does not update build status on github page for dosfstools organization. Probably @andreasbombe as admin of dosfstools organization needs to re-enable/re-approve it.

Here is build link for this pull request: https://travis-ci.org/dosfstools/dosfstools/builds/481711836

Build & test with clang is failing due to some memory problems in dosfstools code itself.

pali:
Now build status on github is updated.

pali:
Now tests are passing.

fstirlitz:
Why `-fwrapv`? Wouldn't actually *catching* overflow errors in CI tests (i.e. `-ftrapv`, or `-fsanitize=signed-integer-overflow`, or outright `-fsanitize=undefined`) be better?

pali:
`-fsanitize` does not work for static builds and it is quite new option in gcc. If you have a way how to improve CI testing, feel free to create a pull request.

PR #116, opened 2019-01-19
Use unsigned long long type instead of uint64_t

pali:
In dosfstools codebase is primary used `unsigned long long` type for 64bit integers and on few places `uint64_t`. So replace `uint64_t` by `unsigned long long` type and fix some casting problems around.


metux:
I don't believe it's a good idea to rely on target specific type sizes. Instead we should carefully check what's really going on here and use proper - also semantically fitting - types (eg. size_t for sizes, etc.).

pali:
@metux: The idea in these changes is that on more places are used arithmetic operation and then results are written by printf-like functions. So I think it is better to use for these purposes `long long` types instead of `int64_t` (or `unsigned int` instead of `off_t` in case result it not passed to function which expects `off_t`).

But I agree that semantically correct types like `size_t` for sizes or `off_t` for file offsets (and therefore also for file sizes) should be used.

If there are not any objections, I would take and merge this pull request.

Issue #110, opened 2019-01-09
"fsck.fat" complains about FSINFO when there is none.


pthfdr-42:
It looks like that "fsck.fat" checks for FSINFO at sector 1, even there is none.
My suggestion is either:
1. Ignore FSINFO check when there are only 2 reserved sectors; or
2. Force a minimum of `-b 2` and `-R 3` when creating FAT32 volumes.

How to reproduce:
Firstly create an FAT FS with alignment. Obviously 62 sectors are wasted:

    root:/u/h/root> mkfs.fat -D 0x00 -f 1 -F 32 -h 0 -I -m /dev/null -M 0xF0 -n TEST -R 2 -s 64 -vv /dev/sdg
    mkfs.fat 4.1 (2017-01-24)
    attribute "partition" not found
    Boot jump code is eb 58
    Using 2 reserved sectors
    Trying with 64 sectors/cluster:
    FAT12: #clu=967090, fatlen=2880, maxclu=4080, limit=4080
    FAT12: too much clusters
    FAT16: #clu=967075, fatlen=3840, maxclu=65520, limit=65520
    FAT16: too much clusters
    FAT16: would be misdetected as FAT12
    FAT32: #clu=967016, fatlen=7616, maxclu=974848, limit=268435440
    Using sector 6 as backup boot sector (0 = none)
    /dev/sdg has 64 heads and 32 sectors per track,
    hidden sectors 0x0000;
    logical sector size is 512,
    using 0xf0 media descriptor, with 61896704 sectors;
    drive number 0x00;
    filesystem has 1 32-bit FAT and 64 sectors per cluster.
    FAT size is 7616 sectors, and provides 967016 clusters.
    There are 64 reserved sectors.
    Volume ID is fce1837e, volume label TEST       .
Then check it. There is no problem:

    root:/u/h/root> fsck.fat -vV /dev/sdg
    fsck.fat 4.1 (2017-01-24)
    Checking we can access the last sector of the filesystem
    Boot sector contents:
    System ID "mkfs.fat"
    Media byte 0xf0 (5.25" or 3.5" HD floppy)
           512 bytes per logical sector
         32768 bytes per cluster
            64 reserved sectors
    First FAT starts at byte 32768 (sector 64)
             1 FATs, 32 bit entries
       3899392 bytes per FAT (= 7616 sectors)
    Root directory start at cluster 2 (arbitrary size)
    Data area starts at byte 3932160 (sector 7680)
[...]

Issue #74, opened 2018-03-07
FAT32 FS Information Sector overwrote by boot sector backup


zhangyoufu:
`mkfs.fat` arguments: `-F 32 -R 2`

the content of sector 0 and 1 are same while the EBPB still points FS Information Sector at sector 1

```
	    backup_boot = (reserved_sectors >= 7) ? 6 :
		(reserved_sectors >= 2) ? reserved_sectors - 1 : 0;
```

pali:
Same problem as desribed in https://github.com/dosfstools/dosfstools/issues/110

FSINFO sector is mandatory for FAT32, but backup boot sector is optional. Therefore when there are only two sectors in reserved FAT area, first must be boot sector and second must be FSINFO sector. And there is not backup boot sector.

So above code is wrong.

PR #117, opened 2019-02-23
mkfs.fat: Fix handling of FAT32 backup and info sectors

pali:
Fixes: https://github.com/dosfstools/dosfstools/issues/74
Fixes: https://github.com/dosfstools/dosfstools/issues/110


saproj:
This pull request introduces more changes than needed to fix #74. One-line change fixes it:
`		(reserved_sectors >= 3) ? reserved_sectors - 1 : 0;`
Other (quite sophisticated) changes to code and a manpage addition can be suggested separately.

Let me point out that this so easily fixable bug #74 has been open for more than a year and since then at least one more person (me) stumbled on it.

pali:
> Is a backup of the info sector actually needed?

Yes, it is required FAT specification (fatgen103.pdf).

So implementation of FAT which is compliance with that specification may handle such FS as in inconsistant state.

If something is written in specification you should really follow it unless you have a very strong argument against (e.g. clean mistake in specification). Otherwise there would be lot of "incompatible" implementations and specification is there to ensure that all implementations would be compatible.

> Does any program look for it?

I do not know.

I just checked that some versions of format.exe on Windows really put there backup of info sector. And because MS's Windows implementation is basically "normative" (FAT is their product) we should follow it there.

> but the info sector is just a cache which can be readily regenerated by scanning the FAT (even though may take some time).

I agree with you. What is the real reason I do not know. You would have to ask authors of specification to get reason.

I can imagine that backup info sector may be used when main one is broken and FAT volume is really huge with small cluster size. This would lead to big FAT table and reconstruction of "cache" in info sector can take too much time. So access to backup is faster.

fstirlitz:
That surprised me, because I have that document and couldn't find it there before. But indeed there it is, page 22. In fact I should have been asking why not also make a copy of the third sector… though I guess it's because `mkfs.fat` puts zeroes there anyway.

> I can imagine that backup info sector may be used when main one is broken and FAT volume is really huge with small cluster size. This would lead to big FAT table and reconstruction of "cache" in info sector can take too much time. So access to backup is faster.

I don't think so. Page 12 of the same specification suggests that the backup is not meant to be continuously kept in sync with the active copy. I also took a look at one FAT32 volume I have (used by a VM), and there is indeed a stale copy of the FSINFO sector there (and a copy of sector 2 as well). So it seems the backup is mostly meant to keep data related to disk layout, which is unlikely to change throughout the lifetime of the file system; given that, the FSINFO backup is unlikely to be relied on anyway.

But spec is spec, I guess.

pali:
> In fact I should have been asking why not also make a copy of the third sector

We both agree that fatgen103.pdf specification is a mess.

First some facts which I think should be clear:

First part of FS is Reserved Region which length is BPB_RsvdSecCnt sectors. Reserved Region starts at sector 0. At sector 0 is stored Boot Sector, at sector 1 is stored FSInfo Sector and at sector 6 is stored Backup Boot Sector.

And then in fatgen103.pdf is written:
>  Note that starting at the BPB_BkBootSec sector is a complete boot record. The Microsoft FAT32 “boot sector” is actually three 512-byte sectors long. There is a copy of all three of these sectors starting at the BPB_BkBootSec sector. A copy of the FSInfo sector is also there

My understanding of this section is following: Boot code must be stored in Reserved Region and must starts at sector 0. Microsoft FAT32 implementation uses boot code with is 3 sectors long. But other implementations are free to have boot code longer or smaller. By complete boot record I understand that boot code which spans more sectors and starts at sector 0. And copy of complete boot record (therefore full boot code) must starts at BPB_BkBootSec sector (6) and continue to next sectors. And into specification wrote that `all three of these sectors` because they referring to their 3 sectors implementation. Other implementations may be larger/smaller. And `A copy of the FSInfo sector is also there` I understand as emphasis of the fact that FSInfo sector is immediately after Boot sector and if you are copying first 3 sectors you would copy also second sector.

As mkfs.fat's boot code is 1 sector long, plus required FSInfo sector which needs to be also there, mkfs.fat `complete boot record` would be only 2 sectors long.

But if you have different understanding of that specification and you thinks that 3rd sector needs to be copied, then we can say it is already done. 3rd sector for mkfs.fat is empty (we do not use it) and because by default Reserved Area is 32 sectors, we have a copy of 3rd sector stored on 8th sector (both are empty, zeroed).

And what is the reason? I really do not know. We can just speculate. Maybe even MS do not know... Another speculation: MS boot code 3 sectors long and second sector (sector 1) is by requirement FSInfo Sector. So their bootloader is stored on sector 0 and sector 2. With possiblity that other implementation (or even MS in future) could have longer bootloader stored on more sectors (therefore on 3, 4, 5...). And to simplify restoration process when sector 0 is broken, just copy from starting sector 6 all boot code (6, 7, 8, ...) to starting sector 0 (0, 1, 2, ...) without need to specially handle and skip sector 7 which is backup copy of FSInfo Sector. And maybe MS did not think that somebody wrote bootloader which fit into one sector and does not need them more...

PR #121, opened 2019-06-05
Use static inline function instead of statement expression

michaelforney:
Not sure if this needs to be inlined, but I made it `inline` to match the existing macro.

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -150,9 +150,9 @@ static void copy_lfn_part(unsigned char *dst, LFN_ENT * lfn)
 /* Convert name part in 'lfn' from unicode to ASCII */
 static inline char *cnv_this_part(LFN_ENT *lfn)
 {
-	unsigned char part_uni[CHARS_PER_LFN*2];
-	copy_lfn_part( part_uni, lfn );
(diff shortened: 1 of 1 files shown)


pali:
@michaelforney can you please rebase this pull request on top of master branch? I fixed Travis CI testing in this (now merged) PR: https://github.com/dosfstools/dosfstools/pull/122

michaelforney:
Done. Looks like it's passing now.

src/lfn.c
```diff
@@ -155,6 +147,14 @@ static void copy_lfn_part(unsigned char *dst, LFN_ENT * lfn)
     memcpy(dst + 22, lfn->name11_12, 4);
 }
 
+/* Convert name part in 'lfn' from unicode to ASCII */
+static inline char *cnv_this_part(LFN_ENT *lfn)
+{
+	unsigned char __part_uni[CHARS_PER_LFN*2];
```
pali: One thing: prefix `__` is reserved for compiler, so it must not be used for user variables in functions.

michaelforney: Done. I also adjusted the whitespace style slightly to match the surrounding code.

PR #124, opened 2019-06-20
configure: Fix iconv check for cross compilation

neheb:
AC_CHECK_LIB is more friendly towards cross-compilation.

Added check for libiconv_open as that can be used when the libc lacks iconv.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,7 +75,8 @@ if test "x$with_udev" != "xno"; then
 		  [true])
 fi
 
-AC_SEARCH_LIBS(iconv_open, iconv)
+AC_CHECK_LIB(iconv, iconv_open)
+AC_CHECK_LIB(iconv, libiconv_open)
 
 # xxd (distributed with vim) is used in the testsuite
 AC_CHECK_PROG([XXD_FOUND], [xxd], [yes])


configure.ac
```diff
@@ -75,7 +75,8 @@ if test "x$with_udev" != "xno"; then
 		  [true])
 fi
 
-AC_SEARCH_LIBS(iconv_open, iconv)
+AC_CHECK_LIB(iconv, iconv_open)
+AC_CHECK_LIB(iconv, libiconv_open)
```
pali: Wrong function name.

neheb: What's the correct one?

pali: `iconv_open`

Look into `src/charconv.c` file, only `iconv_open` function is used, not `libiconv_open`.

neheb: This is true. Except for this gem:

http://git.savannah.gnu.org/cgit/libiconv.git/tree/include/iconv.h.in#n70

libiconv redefines the functions, probably to avoid conflict with the build in libc ones.

pali: But if it is just define for `iconv_open` then check for it is not needed. And if there is no define for `iconv_open` from `libiconv_open` then dosfstools would not use it.

neheb: For reference: here is the build failure: https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/dosfstools/compile.txt

Note that this is also needed under Cygwin for the same reason.

pali: And does not just check for `iconv_open` fixes it? dosfstools uses `iconv_open` therefore in configure check there should be only `iconv_open`. Implementation details (like some function is aliased or redefined or defined as macro) is in important here as configure phase should use same steps for compilation like real project compilation.

neheb: No. AC_CHECK_LIB(iconv, libiconv_open) needs to be there. Otherwise this is the resulting failure.

neheb: For reference: that build failure has this patch already applied: https://github.com/openwrt/packages/blob/master/utils/dosfstools/patches/0002-Switch-to-AC_CHECK_LIB-for-iconv-library-linking.patch

pali: Then `AC_CHECK_LIB` check does not work there correctly... and so patch is not complete.

pali: So what is needed is a test which includes `<iconv.h>` and calls `iconv_open` function like it is used in dosfstools code (in `src/charconv.c` file). If compilation/linking test fails then it should adds external libraries (like adding `-liconv` to LIBS) and try again. If it fails then this configure test must throw fatal error and does not allow compilation of dosfstools project at all -- as dosfstools depends on working `iconv_open` function.

As dosfstools project does not use any `libiconv_open` function, there is absolutely no need to test if `libiconv_open` function is available in system or not. Its check is just a problem in testing system.

neheb: A lot of work for something simple...

pali: Welcome to autohell world where everything is complicated.

As I stated `AC_CHECK_LIB(iconv, libiconv_open)` is wrong as dosfstools does not use `libiconv_open`.

Anyway, I looked at this problem and there is already written M4 macro for iconv, called `AM_ICONV` in `iconv.m4` file. See https://www.gnu.org/software/gettext/manual/html_node/AM_005fICONV.html

I think this is the right way how to support iconv correctly without re-inventing wheel again and again.

neheb: I tried replacing both lines with AM_ICONV. Didn't work. Not sure how to properly use it.

pali: And have you properly set `LIBICONV` as described?

neheb: This is shown in the configure output:

```
checking for iconv... yes
checking for working iconv... guessing yes
checking how to link with libiconv... /home/mangix/devstuff/openwrt/staging_dir/target-arc_arc700_uClibc/usr/lib/libiconv-stub/lib/libiconv.a
```

Which is correct. But in the linker output I do not see -liconv

pali: And have you specified `LIBICONV` for linker?

I tried to create more error-prone solution with own internal CP850 table as fallback when there is either no iconv or CP850 is not supported by iconv. It is in following branch, can you test it?
https://github.com/dosfstools/dosfstools/compare/master...pali:charconv

neheb: Didn't work. Full output: https://gist.github.com/neheb/7b4e29ff2b3e2bcaedf9d5947e405248

neheb: There is no -liconv anywhere. The patch in this pull request adds it, in addition to being simple.

neheb: With this patch: https://gist.github.com/neheb/d6aeb9a3123d75ad8bd0848b1903f514

pali: > There is no -liconv anywhere.

Please check `LIBICONV` and `LTLIBICONV` variables in generated `src/Makefile`.

> The patch in this pull request adds it, in addition to being simple.

It could be simple, but it is incorrect. And incorrect or wrong patches cannot be merged into upstream project.

neheb: > > There is no -liconv anywhere.
> 
> Please check `LIBICONV` and `LTLIBICONV` variables in generated `src/Makefile`.
> 
> > The patch in this pull request adds it, in addition to being simple.
> 
> It could be simple, but it is incorrect. And incorrect or wrong patches cannot be merged into upstream project.

LIBICONV = /home/mangix/devstuff/openwrt/staging_dir/target-arc_arc700_uClibc/usr/lib/libiconv-stub/lib/libiconv.a

LTLIBICONV = -L/home/mangix/devstuff/openwrt/staging_dir/target-arc_arc700_uClibc/usr/lib/libiconv-stub/lib -liconv

pali: That is correct...

neheb: It is. And it's not being passed to the linker. Check the outpit I showed.

pali: In your log is failed compilation of `testdevinfo`, not of mkfs/fsck.

Mhm, this is bug in compile definition for `testdevinfo`, for unknown reason it includes whole mkfs sources even it does not use them. I'm going to fix this bug.

pali: Now it should be fixed in my charconv branch: https://github.com/dosfstools/dosfstools/compare/master...pali:charconv

neheb: Yes. Now it's fixed.

pali: Ok, I rebased this branch, added support to compile without iconv by `--without-iconv` configure flag and created a new pull request: https://github.com/dosfstools/dosfstools/pull/126

neheb:
Works for me. Closing.

configure.ac
pali: I found more problems in charconv code... Also on Travis was non-working autoreconf, so I introduced `autogen.sh`. Pull request https://github.com/dosfstools/dosfstools/pull/126 is now updated. Can you test it? Now also static linked version of dosfstools without iconv should work.

neheb: Tested basic compile. Still works.

PR #129, opened 2019-10-12
Use 'htole' for LE conversion in mkfs.fat.c

davidgumberg:
Switch manual byte conversion to htole in mkfs.fat.c

Corrected #128, still learning git and doing some stupid repo deletion, commit squashing stuff.

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -151,12 +151,12 @@ struct msdos_boot_sector {
     uint8_t boot_jump[3];	/* Boot strap short or near jump */
     uint8_t system_id[8];	/* Name - can be used to special case
 				   partition manager volumes */
-    uint8_t sector_size[2];	/* bytes per logical sector */
+    uint16_t sector_size;	/* bytes per logical sector */
     uint8_t cluster_size;	/* sectors/cluster */
     uint16_t reserved;		/* reserved sectors */
     uint8_t fats;		/* number of FATs */
-    uint8_t dir_entries[2];	/* root directory entries */
-    uint8_t sectors[2];		/* number of sectors */
+    uint16_t dir_entries;	/* root directory entries */
+    uint16_t sectors;		/* number of sectors */
     uint8_t media;		/* media code (unused) */
     uint16_t fat_length;	/* sectors/FAT */
     uint16_t secs_track;	/* sectors per track */
@@ -1136,11 +1136,8 @@ static void setup_tables(void)
         partition[15] = (num_sectors >> 24) & 0xFF;
     }
 
-    bs.sector_size[0] = (char)(sector_size & 0x00ff);
-    bs.sector_size[1] = (char)((sector_size & 0xff00) >> 8);
-
-    bs.dir_entries[0] = (char)(root_dir_entries & 0x00ff);
-    bs.dir_entries[1] = (char)((root_dir_entries & 0xff00) >> 8);
+    bs.sector_size = htole16((uint16_t) sector_size);
+    bs.dir_entries = htole16((uint16_t)root_dir_entries);
 
     if (size_fat == 32) {
 	/* set up additional FAT32 fields */
@@ -1180,12 +1177,10 @@ static void setup_tables(void)
 		   "or higher.\n");
     }
     if (num_sectors >= 65536) {
-	bs.sectors[0] = (char)0;
-	bs.sectors[1] = (char)0;
+	bs.sectors = htole16(0);
 	bs.total_sect = htole32(num_sectors);
     } else {
-	bs.sectors[0] = (char)(num_sectors & 0x00ff);
-	bs.sectors[1] = (char)((num_sectors & 0xff00) >> 8);
+	bs.sectors = htole16((uint16_t) num_sectors);
 	if (!atari_format)
 	    bs.total_sect = htole32(0);
     }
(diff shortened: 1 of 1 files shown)


davidgumberg:
I believe those casts are unnecessary, but I included them in case I don't fully understand something about ``htole``

fstirlitz:
Fields converted in this commit not are aligned at a 16-bit boundary, so it will make code not portable to architectures which disallow unaligned memory accesses.

Honestly, if anything, everything else should be converted the other way around; all the structure fields types' should be changed to `char` arrays, and accessed by serialising/deserialising them byte by byte. This is the only way to deal with binary data in portable C while having full control over endianness and padding; this technique is already used by binutils, for what it's worth. The endian serialisation could be split into `inline static` accessor functions, like [ones I wrote in one of my own projects](https://github.com/fstirlitz/c-bsp/blob/b15e82e08cb02bb1e88113664786c53c3c2e692c/src/lib/util.h#L21-L33). Modern compilers are able to optimise those accessors into inline memory accesses, so they incur no overhead over 'traditional' field packing and/or type punning.

For extra type safety, one could also wrap char arrays in semi-opaque structs:
```c
typedef struct { char data[4]; } u32le_t;

inline static uint32_t get_u32le(const u32le_t p[static 1]) { /* ... */ }
inline static void put_u32le(u32le_t p[static 1], uint32_t value) { /* ... */ }
```
This way the compiler can generate warnings in case the wrong accessor function is used by accident.

pali:
@fstirlitz: You are rigth that it is a problem for unaligned access to structure members. Bigger issue is that unaligned access is allowed on x86 (it is just slower as aligned access) therefore in most cases people would not see that code is broken (= it works on my machine which is x86).

But you forgot one important note. These structures are marked with GCC specific attribute `__attribute__ ((packed))` which allows unaligned access to structure members. This GCC extention generates for all architectures code which workaround problems with unaligned access. So we can safelly write it.

Yes, this means that code is less portable, it needs compiler with these GCC extensions. But it was there for a long time and basically every used C compiler on Linux supports it. And because dosfstools project is for Linux, it does not care about compilers on other systems.

pali:
Moreover,

> ```c
> typedef struct { char data[4]; } u32le_t;
> ```

this code is broken. Compiler is free to put any padding after `data` member. So `u32le_t` does not have to be 32bit long as its name suggest.

fstirlitz:
> This GCC extention generates for all architectures code which workaround problems with unaligned access.

I didn't know that attribute had that effect. Does it also cover the case of taking an address of a packed field and passing it around as a pointer? Because if yes, then it follows that it should treat every pointer dereference as potential unaligned memory access, which I imagine would incur quite an overhead. (Or is `foo.bar` fine, but `*&foo.bar` not? Or is that fine too, but `({ __auto_type pbar = &foo.bar; *pbar; })` not?)

> Compiler is free to put any padding after `data` member. So `u32le_t` does not have to be 32bit long as its name suggest.

Strictly speaking, the standard doesn't restrict in any way what kind of padding appears at the end of a structure (ISO/IEC 9899:1999 §6.7.2.1, par. 15), but if we're playing language-lawyer games then it also doesn't restrict what kind of padding appears *between* structure members (op. cit., par. 13). So
```c
struct foo {
    uint16_t x;
    uint16_t y;
}
```
is allowed to have padding between `x` and `y`. But I haven't seen a compiler that does either of these things; I think most of them only insert padding bytes to satisfy alignment requirements.

If you're sufficiently paranoid about it, you should probably at least write compile-time assertions (either `_Static_assert`, or good ol' `extern char __dummy[(cond) ? 1 : -1];`) that the sizes of your structures are correct (and maybe even check the offset of each individual structure member). But you should probably write those anyway, and not (just) for the sake of spiteful compilers, but to catch (at least some) honest coding mistakes in structure layout declarations. It did help me a couple times at least.

pali:
Now I tried to find exact information about packed strctures on gcc webpages, but I was not able...

From what I remember is that reading and storing from structure is safe. Taking a pointer to member is ok, but you cannot dereference it as it would be unaligned pointer. Some people complained that on some architectures gcc generates too big code for any operation with packed structure and therefore packed attribute should be avoided where it is possible. I guess that too big code "simulates" copy of data to temporary aligned address and then process data from that temporary address.

About padding between members, you are right. The only place where padding cannot be done by compiler is before first member. It is guaranteed that pointer address to the first member is same as address to the whole structure. I have not did investigation of existing compilers in this area, so I do not know about behavior.

Static assert (either via negative array size or via C11 keyword) is a good idea. It basically declares "constraint" supported by some other languages.

fstirlitz:
Actually, it's not that hard to find: there's [an SO question](https://stackoverflow.com/a/8568441), which leads to [GCC PR #51628](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628). And apparently direct access to packed fields is fine, and so is immediately dereferencing a just-formed pointer (`*&`), but passing that pointer along further is not.

So, other than not being portable to compilers without `__attribute__((packed))`, this code seems fine after all.

pali:
`__attribute__((packed))` is already used and needed for mkfs.fat, so this pull request does not decrease portability of code. And I think that code is correct too.

davidgumberg:
Sorry fstirlitz, I failed to understand your first comment. Let's say the struct wasn't packed. What would be the problem with this PR? Where is the unaligned memory access?

fstirlitz:
`sector_size` is at offset 11 in the boot sector, `dir_entries` at offset 17 and `sectors` at offset 19. If the `packed` attribute weren't there, the compiler would assume a byte of padding before each but the last, which is not how the boot sector is actually laid out. With the attribute, the compiler generates correct code to access it at that address, even on targets that don't support direct unaligned accesses; I assumed this wasn't the case.

pali:
@davidgumberg: Could you rebase this patch on top of master branch? Or should I do it and merge it?

davidgumberg:
I will rebase later tonight, I have not looked in a long time but I think I missed a spot in this PR

pali:
PR is OK, just needs to be rebased on top of master branch as there are some git conflicts.

pali:
@davidgumberg: Could you please rebase this pull request?

pali:
ping

davidgumberg:
@pali Rebased! Sorry, about that, must have fallen asleep.

pali:
Merged now, thanks!

PR #131, opened 2019-10-14
Fix dynamically linked PPC & ARM travis builds

### LINKED ISSUE #130: Dynamically Linked ARM and PPC Travis builds are failing
https://travis-ci.org/davidgumberg/dosfstools

I was losing hair trying to figure out why #129 was failing build, don't know much about build stuff, but travis segfaults when running the tests on my master branch (even with this one) with the dynamic link ARM and PPC builds, even though it says this branch is passing build so something must have changed with the versions of the dynamic libraries since the build was last done 3 months ago I'm guessing.

davidgumberg:
I won't pretend to understand why, but getting iconv config from the platform's libc causes segfault.

This PR Fixes #130

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,10 @@ matrix:
             - libc6-powerpc-cross
             - qemu-user
       before_script:
-        - wget http://ports.ubuntu.com/pool/main/e/eglibc/libc6_2.19-0ubuntu6_powerpc.deb -O /tmp/libc6_powerpc.deb
+        - VERSION=`dpkg-query -W -f='${Version}\n' libc6-powerpc-cross | head -1 | sed 's/\(ubuntu[0-9]*\).*/\1/'`
+        - SOURCE=`dpkg-query -W -f='${Source}\n' libc6 | head -1`
+        - PREFIX=`echo "$SOURCE" | sed 's/^\(.\).*/\1/'`
+        - wget "http://ports.ubuntu.com/pool/main/$PREFIX/$SOURCE/libc6_${VERSION}_powerpc.deb" -O /tmp/libc6_powerpc.deb
         - dpkg -x /tmp/libc6_powerpc.deb /tmp/libc6_powerpc
         - export GCONV_PATH=/tmp/libc6_powerpc/usr/lib/powerpc-linux-gnu/gconv
     - compiler: arm-linux-gnueabi-gcc
@@ -71,7 +74,10 @@ matrix:
             - libc6-armel-cross
             - qemu-user
       before_script:
-        - wget http://ports.ubuntu.com/pool/main/e/eglibc/libc6-armel_2.19-0ubuntu6_armhf.deb -O /tmp/libc6-armel_armhf.deb
+        - VERSION=`dpkg-query -W -f='${Version}\n' libc6-armel-cross | head -1 | sed 's/\(ubuntu[0-9]*\).*/\1/'`
+        - SOURCE=`dpkg-query -W -f='${Source}\n' libc6 | head -1`
+        - PREFIX=`echo "$SOURCE" | sed 's/^\(.\).*/\1/'`
+        - wget "http://ports.ubuntu.com/pool/main/$PREFIX/$SOURCE/libc6-armel_${VERSION}_armhf.deb" -O /tmp/libc6-armel_armhf.deb
         - dpkg -x /tmp/libc6-armel_armhf.deb /tmp/libc6-armel_armhf
         - export GCONV_PATH=/tmp/libc6-armel_armhf/usr/lib/arm-linux-gnueabi/gconv
 


pali:
Into master branch I pushed commit which should fix collecting errors from fatlabel tests. Please rebase your pull request on top of master branch, to see if your change really fixed tests. I bet that no.

davidgumberg:
Is this an upstream issue with AM_ICONV macro?

pali:
There are problems with `autoreconf` related to gettext and `AM_ICONV`, see:
https://github.com/dosfstools/dosfstools/commit/b657ede4abd91a30c4cd430758cf42d9b8bef1c3

But this one is not related to it. This is rather Ubuntu packaging problem. GNU libc's implement of iconv is modular and encodings are implemented in external libraries/plugins. iconv itself works, but is useless if you do not have needed "plugins".

For this reason I implemented fallback mechanism for dosfstools and implemented internal encoding support for CP850 encoding. This is used when either iconv library is not available at compile time or when GNU libc's implementation is used, but at runtime is missing needed plugin for CP850.

And that test is failing because it need to use iconv library and not my fallback implementation.

So to fixing it, it is needed to grab needed libraries/plugins from Ubuntu and put them into system. I guess that plugin version must match libc version (where is iconv implemented).

pali:
Try this one:
```diff
diff --git a/.travis.yml b/.travis.yml
index ae6aaab..baec207 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,11 @@ matrix:
             - libc6-powerpc-cross
             - qemu-user
       before_script:
-        - wget http://ports.ubuntu.com/pool/main/e/eglibc/libc6_2.19-0ubuntu6_powerpc.deb -O /tmp/libc6_powerpc.deb
+        - PACKAGE="libc6-powerpc-cross"
+        - VERSION=`dpkg-query -W -f='${Version}\n' "$PACKAGE" | head -1 | sed 's/\(ubuntu[0-9]*\).*/\1/'`
+        - SOURCE=`dpkg-query -W -f='${Source}\n' "$PACKAGE" | head -1`
+        - PREFIX=`echo "$SOURCE" | sed 's/^\(.\).*/\1/'`
+        - wget "http://ports.ubuntu.com/pool/main/$PREFIX/$SOURCE/libc6_${VERSION}_powerpc.deb" -O /tmp/libc6_powerpc.deb
         - dpkg -x /tmp/libc6_powerpc.deb /tmp/libc6_powerpc
         - export GCONV_PATH=/tmp/libc6_powerpc/usr/lib/powerpc-linux-gnu/gconv
     - compiler: arm-linux-gnueabi-gcc
@@ -71,7 +75,11 @@ matrix:
             - libc6-armel-cross
             - qemu-user
       before_script:
-        - wget http://ports.ubuntu.com/pool/main/e/eglibc/libc6-armel_2.19-0ubuntu6_armhf.deb -O /tmp/libc6-armel_armhf.deb
+        - PACKAGE="libc6-armel-cross"
+        - VERSION=`dpkg-query -W -f='${Version}\n' "$PACKAGE" | head -1 | sed 's/\(ubuntu[0-9]*\).*/\1/'`
+        - SOURCE=`dpkg-query -W -f='${Source}\n' "$PACKAGE" | head -1`
+        - PREFIX=`echo "$SOURCE" | sed 's/^\(.\).*/\1/'`
+        - wget "http://ports.ubuntu.com/pool/main/$PREFIX/$SOURCE/libc6-armel_${VERSION}_armhf.deb" -O /tmp/libc6-armel_armhf.deb
         - dpkg -x /tmp/libc6-armel_armhf.deb /tmp/libc6-armel_armhf
         - export GCONV_PATH=/tmp/libc6-armel_armhf/usr/lib/arm-linux-gnueabi/gconv
 
```
Hardcoded version is for sure a problem.

pali:
(I updated patch, `Source` should be read from `libc6` package)

davidgumberg:
Holy shit, nice!

So was the problem the version or that it was getting eglibc?

pali:
As I wrote in https://github.com/dosfstools/dosfstools/issues/130 Travis updated Ubuntu from Trusty to Xenial and so also libc6 (and its iconv implementation) was updated. So for correct working it needed new version of iconv libraries and plugins.

Problem was that job was trying to use old iconv libraries with new (incompatible) libc6 implementation. So new version of DEB package was needed to download (plus its location was changed from eglibc to glibc). In that proposed patch I tried to write code which downloads correct version of libc (instead of hardcoded one) so it would be more error prone.

And seems it is working now!

I would suggest to remove those empty lines between `PREFIX=` and `wget` as I'm not fully sure how YAML specification define such lists...

davidgumberg:
Looks good, nice!

pali:
Great! Now I merged it.

PR #133, opened 2020-01-16
More fixes for charconv code

pali:
Fix fallback translit code and add internal CP850 translit table. Fix buffer overflows by specifying output buffer size. Fix usage of iconv() function and use ICONV_CONST autoconf macro.

### DIFF
--- a/src/charconv.c
+++ b/src/charconv.c
@@ -72,6 +72,10 @@ static int wchar_string_to_cp850_string(char *out, const wchar_t *in, unsigned i
             return 0;
         }
     }
+    if (in[i]) {
+        fprintf(stderr, "Cannot convert input string to 'CP850': String is too long\n");
+        return 0;
+    }
     out[i] = 0;
     return 1;
 }
@@ -204,13 +208,12 @@ int local_string_to_dos_string(char *out, char *in, unsigned int out_size)
         return local_string_to_cp850_string(out, in, out_size);
     ret = iconv(local_to_dos, &pin, &bytes_in, &pout, &bytes_out);
     if (ret == (size_t)-1) {
-        fprintf(stderr, "Cannot convert input sequence '\\x%.02hhX' from codeset '%s' to 'CP%d': %s\n",
-                *pin, nl_langinfo(CODESET), used_codepage, strerror(errno));
-        return 0;
-    }
-    if (bytes_in != 0) {
-        fprintf(stderr, "Cannot convert input string '%s' to 'CP%d': String is too long\n",
-                in, used_codepage);
+        if (errno == E2BIG)
+            fprintf(stderr, "Cannot convert input string '%s' to 'CP%d': String is too long\n",
+                    in, used_codepage);
+        else
+            fprintf(stderr, "Cannot convert input sequence '\\x%.02hhX' from codeset '%s' to 'CP%d': %s\n",
+                    *pin, nl_langinfo(CODESET), used_codepage, strerror(errno));
         return 0;
     }
     out[out_size-1-bytes_out] = 0;
--- a/src/charconv.c
+++ b/src/charconv.c
@@ -188,7 +188,7 @@ int dos_char_to_printable(char **p, unsigned char c, unsigned int out_size)
 	return 0;
     if (internal_cp850)
         return cp850_char_to_printable(p, c, out_size);
-    return iconv(dos_to_local, &pin, &bytes_in, p, &bytes_out) != -1;
+    return iconv(dos_to_local, &pin, &bytes_in, p, &bytes_out) != (size_t)-1;
 }
 
 int local_string_to_dos_string(char *out, char *in, unsigned int out_size)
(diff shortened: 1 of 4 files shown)


pali:
@fstirlitz could you look at changes in this pull request?

fstirlitz:
...why me? (Not that I couldn't, but...)

pali:
You have already looked at some other changes, so I thought that you could look at this one too.

fstirlitz:
On first glance, nothing seems particularly egregious. ASCII transliteration for CP850 characters looks interesting, but I worry a bit that a user who is presented with those might be unable to distinguish, say, `Ü.TXT` from `U.TXT`. (Although it's not on a FAT volume, I do have a project containing files with such similar short names, a scenario where this matters is not entirely unreasonable.) It would be better if the program showed a warning about inconvertible characters, and/or that transliterations were clearly visually marked as distinct from literal representations of characters. (For example, one may wrap them in `[` `]`, which are otherwise disallowed in SFNs, although not in LFNs.) But then again, that was already a problem with `iconv` and `//TRANSLIT`. (Never mind that `//TRANSLIT` is not portable to non-GNU implementations of `iconv`.)

Were I to do it, I'd consider ditching iconv and simply embedding mapping tables from DOS code page to Unicode code points, and then converting those to the native system charset using `wcrtomb` or `wcsrtombs`. There aren't that many of them anyway, the conversion is trivial (code pages are not stateful encodings), and (at least some of) the tables can be readily generated from the [data files on unicode.org](https://www.unicode.org/Public/MAPPINGS/VENDORS/MICSFT/PC/) (874 is not a DOS code page, the others are); licensing should not be a problem, since those are available under [the license for Unicode data files](https://www.unicode.org/copyright.html). This way I'd be fully in control over which code pages are supported (I don't think iconv is guaranteed to support any particular encoding, or all combinations of encodings) and whether conversions are lossy or not.

Also worth considering: a mode that puts no Unicode interpretation whatsoever on non-ASCII characters and always displays them as escapes; and a mode that skips all conversions and simply assumes that the FAT code page and system locale are congruent (if the DJGPP port is ever revived, it would come handy there).

src/charconv.c
```diff
@@ -72,6 +72,10 @@ static int wchar_string_to_cp850_string(char *out, const wchar_t *in, unsigned i
             return 0;
         }
     }
+    if (in[i]) {
+        fprintf(stderr, "Cannot convert input string '%ls' to 'CP850': String is too long\n", in);
```
fstirlitz: Maybe you should reuse the `CP%d` string here, and append hardcoded 850 as an argument (or actually `DEFAULT_DOS_CODEPAGE`).

pali:
> but I worry a bit that a user who is presented with those might be unable to distinguish, say, `Ü.TXT` from `U.TXT`

Users who are able to touch `Ü.TXT` file names needs to have locale/codeset with support for umlauts. So displaying these characters would work fine.

> Never mind that `//TRANSLIT` is not portable to non-GNU implementations of `iconv`.

I know and because of it, there is a third commit which tries to use iconv also without `//TRANSLIT`.

And still if iconv conversion fails then there is another fallback code which tries to print it as escaped character or character is replaced by underline.

> I'd consider ditching iconv and simply embedding mapping tables from DOS code page to Unicode code points

I was thinking about it and I rather decided to embed just CP850 code page which is default one. And also iconv is now only optional dependency, so if you compile dosfstools without it, there is still that embedded CP850 table. And in this pull request this table is extended with translit option.

I think that current code is robust enough.

But if somebody has needs for dosfstools without iconv dependency, but with different codepage as CP850, I'm not against accepting a patch which will do it.

> a mode that skips all conversions and simply assumes that the FAT code page and system locale are congruent

This does not work as every non-Windows system does not use UTF-16 as system code page. Do not forget that all VFAT filenames are stored in UTF-16. DJGPP port is for DOS and DOS does not use UTF-16 as default code page, it supports only 8bit code pages.

src/charconv.c
pali: Ok, no problem, I can change it.

pali: Hm... looking at it and I think it does not make sense to replace it with `DEFAULT_DOS_CODEPAGE`. This function is named `wchar_string_to_cp850_string` (and not `to_default_string`) so it should print 850.

fstirlitz: Well, you already merged this, but it might make some sense to make up your mind on the extent to which you want to hardcode the choice of the code page. Either make it ‘internally generic’ (so that you only need to swap the code page table and number to change it) or remove the `DEFAULT_DOS_CODEPAGE` macro.

I’d suggest the former; it should make ditching iconv later (if you so choose) a bit easier.

fstirlitz:
> Users who are able to touch `Ü.TXT` file names

This isn’t the same as users who have access to a file system with such names and might want to check or repair it.

src/charconv.c
pali: charconv contains fallback cp850 codepage which is ised when iconv is not available or when dosfstools was compiled without iconv support. This metioned function implements cp850 conversion, not generic _default_ conversion. So error message in function is correct.

pali:
If users want to repair unicode related stuff they should have unicode aware system. On non-unicode aware or 7bit ASCII system is `U` is the best aproximation of `Ü`. The whole thing of internal fallback cp850 table is ability to show at at least something on systems where iconv is not available. Or where iconv is available but particular dos cp plugins not.

PR #142, opened 2020-08-03
mkfs: fsync() the block device before exiting

poettering:
Let's make sure to sync the block device, before exiting. Otherwise, if
we operate on a loop device and people issue "losetup -d" right after
this command finishes our in-flight writes might never hit the disk.

(This took 4h of my life for me to figure out. The mkfs for all other
file systems I tested appear to synchronize when done, hence fat should
do too.)

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -1795,5 +1795,10 @@ int main(int argc, char **argv)
 
     write_tables();		/* Write the filesystem tables away! */
 
+    /* Let's make sure to sync the block device. Otherwise, if we operate on a loop device and people issue
+     * "losetup -d" right after this command finishes our in-flight writes might never hit the disk */
+    if (fsync(dev) < 0)
+        pdie("unable to synchronize %s", device_name);
+
     exit(0);			/* Terminate with no errors! */
 }


src/mkfs.fat.c
```diff
@@ -1795,5 +1795,14 @@ int main(int argc, char **argv)
 
     write_tables();		/* Write the filesystem tables away! */
 
+    if (fsync(dev) < 0) {
+        /* Let's make sure to sync the block device. Otherwise, if we operate on a loop device and people issue
+         * "losetup -d" right after this command finishes our in-flight writes might never hit the disk */
+
+        fprintf(stderr, "%s: unable to synchronize %s: %s\n", program_name,
+                device_name, strerror(errno));
+        exit(1);
```
pali, line 1804: Please use provided `pdie` function for this purpose.

poettering:
Force pushed a new version, using `pdie()`.

Note that I just copied how code a few lines up printed its error. Also note that the only occasion where `pdie()` was so far used in `mkfs.fat.c` actually uses it incorrectly. `pdie()` outputs `strerror(errno)` implicitly after all, but the error condition where it's used is not based on `errno`, i.e. it will output the wrong error...

src/mkfs.fat.c
```diff
@@ -1795,5 +1795,13 @@ int main(int argc, char **argv)
 
     write_tables();		/* Write the filesystem tables away! */
 
+    if (fsync(dev) < 0) {
+        /* Let's make sure to sync the block device. Otherwise, if we operate on a loop device and people issue
+         * "losetup -d" right after this command finishes our in-flight writes might never hit the disk */
+
+        pdie("unable to synchronize %s", device_name);
+        exit(1);
```
pali, line 1803: `pdie` does not return, it calls `exit`, so do not call `exit` function in your patch second time. That's why I suggested you to use this `pdie` function to reduce copying code; look at `pdie` implementation.

pali:
> Also note that the only occasion where `pdie()` was so far used in `mkfs.fat.c` actually uses it incorrectly. `pdie()` outputs `strerror(errno)` implicitly after all, but the error condition where it's used is not based on `errno`, i.e. it will output the wrong error...

I guess you mean that `vfprintf()` in `pdie()` could reset `errno` or `vfprintf()` itself can fail and change `errno`. It is highly unexpected and if such thing happen there are probably more important problems... But I agree that for correctness it should be fixed.

poettering:
> I guess you mean that `vfprintf()` in `pdie()` could reset `errno` or `vfprintf()` itself can fail and change `errno`. It is highly unexpected and if such thing happen there are probably more important problems... But I agree that for correctness it should be fixed.

No, i mean this:

https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c#L680

`mbstowcs()` doesn't set `errno` afaik, and hence `pdie()` will show the `errno` that was previously set, i.e. not the error of `mbstowcs()`.

pali:
> `mbstowcs()` doesn't set `errno` afaik

https://pubs.opengroup.org/onlinepubs/9699919799/functions/mbstowcs.html
**If an invalid character is encountered, `mbstowcs()` shall return `(size_t)-1` and shall set `errno` to indicate the error.**

pali:
> `mbstowcs()` doesn't set `errno` afaik

I have checked it and...

glibc sets errno: https://sourceware.org/git/?p=glibc.git;a=blob;f=wcsmbs/mbsrtowcs_l.c;h=8eab91f047c174df97b27bb02b4f58b34d7527e5#l159

and musl sets errno too: https://git.musl-libc.org/cgit/musl/tree/src/multibyte/mbsrtowcs.c?h=v1.2.0#n117

> `pdie()` was so far used in `mkfs.fat.c` actually uses it incorrectly.

So, it was used correctly. Please read relevant documentation and implementation prior making such statement. Thanks.

poettering:
Hmm, I read the Linux man pages. Where this is not mentioned, sorry.

https://man7.org/linux/man-pages/man3/mbstowcs.3.html

pali (approved):

Issue #159, opened 2021-02-01
Cannot build v4.2 on macOS: fatal error: 'sys/sysmacros.h' file not found


i0ntempest:
device_info.c and blkdev.c includes this file and build fails on these two files. Removing the two includes and the build completes successfully, and binaries seem to run as expected.

PR #158, opened 2021-01-31
fix: patch for osx build

chenrui333:
```
device_info.c:27:10: fatal error: 'sys/sysmacros.h' file not found
#include <sys/sysmacros.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

relates to https://github.com/Homebrew/homebrew-core/pull/70104

PR #166, opened 2021-08-20
Fix building on macos

i0ntempest:
This is the same as #158 but based on master and includes the required change of configure.ac (https://github.com/dosfstools/dosfstools/pull/158#issuecomment-770943078) to avoid build failure on Debian 10

Fixes: #158
Fixes: #159

### DIFF
--- a/src/blkdev/blkdev.c
+++ b/src/blkdev/blkdev.c
@@ -7,11 +7,14 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/ioctl.h>
-#include <sys/sysmacros.h>
 #include <unistd.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <limits.h>
+
+#ifdef HAVE_SYS_SYSMACROS_H
+#include <sys/sysmacros.h>
+#endif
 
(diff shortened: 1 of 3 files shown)


src/device_info.c
```diff
@@ -24,7 +24,10 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/ioctl.h>
-#include <sys/sysmacros.h>
+
+#ifdef HAVE_SYS_SYSMACROS_H
+# include <sys/sysmacros.h>
```
pali, line 29: Just remove space at beginning to following style in this file.

src/blkdev/blkdev.c
```diff
@@ -7,7 +7,9 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/ioctl.h>
-#include <sys/sysmacros.h>
+#ifdef HAVE_SYS_SYSMACROS_H
+# include <sys/sysmacros.h>
+#endif
```
pali, line 12: Put whole #ifdef section after limits.h where are other includes guarded by #ifdef.

pali:
Thanks, merged!

Issue #172, opened 2022-01-21
fsck.fat destroys volume label from old mkdosfs


fabled:
Old mkdosfs created volume labels which are zero padded. The new fsck.fat notices that it's not space padded and considers it invalid and destroys the label. Upgrading dosfatools and running fsck caused bricking my devices which mounted partitions by label.
Would it be possible for fsck to notice if the label is just zero padded and fix the padding instead of losing the volume label?

pali:
Hello @fabled! Could you please provide example of such problematic fat image? It is possible to upload GZIPed fat images into github as attachments.

fabled:
On further examination it seems the `mkfs.vfat` used to create the filesystem was actually from busybox which seems to have a bug in label creation. Find attached empty image created with:
```
truncate -s 50M test.img
busybox mkfs.vfat test.img  -n TEST
gzip test.img
```
[test.img.gz](https://github.com/dosfstools/dosfstools/files/7912532/test.img.gz)

On this image, I get:
```
$ fsck.vfat -a test.img 
fsck.fat 4.2 (2021-01-31)
Volume label 'TEST' stored in root directory is not valid.
  Auto-removing label.

*** Filesystem was changed ***
Writing changes.
test.img: 1 files, 1/100818 clusters
```

pali:
Ah :-( Please report this issue to the busybox, so they fix it.

It is pity that crappy tools like this one still exists and therefore it is needed to add hacks to fsck tool.

I will try to implement some hack/workaround in fsck.

I would suggest to not use labels for automounting FAT partitions. There were many issues with labels even in "mount" tool (or rather in its libblkid library) and also in dosfstools itself. See fatlabel manpage for details.

Automounting based on FAT UUID is better option.

fabled:
> Ah :-( Please report this issue to the busybox, so they fix it.

Sent a patch for this one:
http://lists.busybox.net/pipermail/busybox/2022-January/089416.html

pali:
This patch is not enough. It will still produce broken label if it starts with byte 0xE5 which is char Õ in English (United Kingdom) locale.

pali:
After `memcpy(de->name, volume_label, sizeof(de->name));` line you need something like this:
```c
if ((unsigned char)de->name[0] == 0xE5)
  de->name[0] = 0x05;
```

pali:
[...]

PR #184, opened 2022-10-19
Add flag to skip validate_volume_label

pzmarzly:
Extra validations were introduced in ca549534762c, let's have an option to skip them. All tools I tested can deal with invalid characters correctly.

Tested with:

    $ sudo src/fatlabel /dev/vda1 A_B
    $ sudo src/fatlabel /dev/vda1 A+B
    fatlabel: labels with characters *?.,;:/\|+=<>[]" are not allowed
    $ sudo src/fatlabel /dev/vda1 A+B --skip-name-validation
    $ sudo mount /dev/vda1 /boot/efi
    $ sudo blkid -c /dev/null | grep vda1
    /dev/vda1: SEC_TYPE="msdos" LABEL_FATBOOT="A+B" LABEL="A+B" \
    UUID="ECFC-989A" BLOCK_SIZE="512" TYPE="vfat" \
    PARTLABEL="EFI System Partition" \
    PARTUUID="5c481284-4af4-437f-93b6-6fb2721a75d2"

    $ fallocate -l 1G ~/tmp.img
    $ src/mkfs.fat -n A+B ~/tmp.img
    mkfs.fat 4.2+git (2021-01-31)
    mkfs.fat: Labels with characters *?.,;:/\|+=<>[]" are not allowed
    $ src/mkfs.fat -n A+B ~/tmp.img --skip-name-validation
    mkfs.fat 4.2+git (2021-01-31)
    $ file ~/tmp.img
    /home/pzmarzly.linux/tmp.img: DOS/MBR boot sector, \
    code offset 0x58+2, OEM-ID "mkfs.fat", sectors/cluster 8, \
    Media descriptor 0xf8, sectors/track 63, heads 64, \
    sectors 2097144 (volumes > 32 MB), FAT (32 bit), sectors/FAT 2048, \
    serial number 0x5ccb07ce, label: "A+B        "

### DIFF
--- a/manpages/fatlabel.8.in
+++ b/manpages/fatlabel.8.in
@@ -51,6 +51,8 @@ Remove label in label mode or generate new ID in volume ID mode.
 .IP "\fB-c\fP \fIPAGE\fP, \fB\-\-codepage\fP=\fIPAGE\fP" 4
 Use DOS codepage \fIPAGE\fP to encode/decode label.
 By default codepage 850 is used.
+.IP "\fB\-\-skip-name-validation\fP" 4
+Do not validate filesystem label.
 .IP "\fB\-h\fP, \fB\-\-help\fP" 4
 Display a help message and terminate.
 .IP "\fB\-V\fP, \fB\-\-version\fP" 4
--- a/manpages/mkfs.fat.8.in
+++ b/manpages/mkfs.fat.8.in
@@ -220,6 +220,8 @@ volume ID and creation time.
 Multiple runs of \fBmkfs.fat\fP on the same device create identical results
 with this option.
 Its main purpose is testing \fBmkfs.fat\fP.
+.IP "\fB\-\-skip-name-validation\fP" 4
+Do not validate filesystem label.
 .\" ----------------------------------------------------------------------------
 .SH BUGS
 \fBmkfs.fat\fP can not create boot-able filesystems.
--- a/src/common.c
+++ b/src/common.c
@@ -43,6 +43,7 @@
 int interactive;
 int write_immed;
 int atari_format;
+int skip_name_validation;
 const char *program_name;
 
 
@@ -340,6 +341,9 @@ int validate_volume_label(char *doslabel)
     int i;
     int ret = 0;
     wchar_t wlabel[12];
+
+    if (skip_name_validation)
+        return 0;
 
     if (dos_string_to_wchar_string(wlabel, doslabel, sizeof(wlabel))) {
         for (i = 0; wlabel[i]; i++) {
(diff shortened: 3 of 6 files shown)


pali:
> Extra validations were introduced in ca549534762c, let's have an option to skip them.

Skipping is against FAT specification defined by Microsoft. Such characters are invalid and must not be present in valid FAT filesystem.

> All tools I tested can deal with invalid characters correctly.

Then you did incomplete research. The first and the main VFAT implementation (Win9x) does not accept such characters and its chkdsk tool replaces all those invalid characters by hyphens.

And I think that older non-VFAT DOS implementation would not handle them too.

There may be many more either VFAT or non-VFAT implementation which do not have to handle filenames with such invalid characters. Or that their repair tools remove/replaces invalid characters.

If one or few tested tools accepts broken filesystems with such characters, it does not mean that it is accepted to start generating such broken filesystems. See also Postel's Law.

So NAK, closing.

davide125:
That's fair, thanks for reviewing @pali. For future readers, and to add more context on why we wanted this: historically, dosfstools allowed one to create filesystems with labels such as `/boot`. These work fine on Linux, and we hadn't realized they were actually invalid until https://github.com/dosfstools/dosfstools/commit/ca549534762c794e5a4cdc36143d17635d196c6a . The idea behind having a flag to skip validation was to allow a stopgap solution while we update our existing tooling and configuration to stop relying on the invalid labels (as that'll take a while). I suspect we're not be the only ones with this issue, and other environments might appreciate the ability to relax this check as a temporary measure.

pali:
If some older version of software was too benevolent and allowed to create an invalid filesystem then it does not mean that new version has too still be able to generate invalid filesystem. Bugs must be fixed and not workarounded. If your tool is trying to generate invalid filesystem then you should fix your tools and environment around. Not trying to convince other projects (like this) to workaround your bugs, just because you have not fixed them yet in your software.

Anyway mentioned commit is there for more than 4 years. What would mean temporary workaround? To stay there another 4*2=8 years? Really I'm not going to accept any temporary solution for other projects as for example in related issue #172 I was waiting half of year and issue in external project was not fixed yet. I see that there is zero interest to fix issues in other projects if somebody allows temporary workaround in main project.

There are many many bugs in dosfstools which must be fixed. There is NO reason to have _forked_ FAT filesystem incompatible with specification and MS implementations. Really, the primary purpose is to have fully compatible implementation with MS.

pali:
For something like _temporary_ solution in commit bd7924ebc8e9f8d772f77ab5345ab78057e00207 I added options into interactive repair mode to skip modifying label. All other repair code has an option to skip doing reparation, so it makes sense to have it also for label code. I hope that this can be useful for a _temporary_ solution.

PR #188, opened 2023-01-03
Support multiple codepage command line arguments.

ferivoz:
Use last given codepage argument instead of first to comply with other command line arguments.

Proof of Concept:

1. Prepare a filesystem with label ° in codepage 1252
```
$ dd if=/dev/zero of=poc.iso bs=512 count=100
$ mkfs.fat -c 1252 -n "°" poc.iso
```

2. Query the label
```
$ fatlabel -c 1252 poc.iso
$ fatlabel -c 850 -c 1252 poc.iso
```

The output should be identical.

Signed-off-by: Samanta Navarro <ferivoz@riseup.net>

### DIFF
--- a/src/fatlabel.c
+++ b/src/fatlabel.c
@@ -221,7 +221,7 @@ int main(int argc, char *argv[])
     char *device = NULL;
     char *new = NULL;
     char *tmp;
-    long codepage;
+    long codepage = -1;
     int c;
 
     check_atari();
@@ -243,8 +243,6 @@ int main(int argc, char *argv[])
 		    fprintf(stderr, "Invalid codepage : %s\n", optarg);
 		    usage(argv[0], 1, 0);
 		}
-		if (!set_dos_codepage(codepage))
-		    usage(argv[0], 1, 0);
 		break;
 
 	    case 'V':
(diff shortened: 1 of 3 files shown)


src/fatlabel.c
```diff
@@ -267,8 +265,8 @@ int main(int argc, char *argv[])
 	}
     }
 
-    if (!set_dos_codepage(-1))	/* set default codepage if none was given in command line */
-        exit(1);
+    if (!set_dos_codepage(codepage))
+        codepage == -1 ? exit(1) : usage(argv[0], 1, 0);
```
pali, line 269: Ternary operator here (when both functions are void and noreturn) looks very ugly. I think it does not bring any value to call usage() here as user already supplied parseable codepage number which operating system does not support (= this is not a bug nor error in dosfstools) and so showing usage does not help. I would suggest to just unconditionally call exit().

pali (changes requested):
Hello! Change looks good, it makes sense that last option wins. Just see a small note about usage().

src/fatlabel.c
ferivoz: I have adjusted my changes accordingly. It also sounds much better with your reasoning. I did not adjust the observable results of the programs to make sure that everything works as before. But I prefer this idea as well.

pali (approved):

pali:
Thanks, merged!

PR #194, opened 2023-03-15
mkfs.fat: nicer bad block messages

ferivoz:
I have reworked the bad block processing in mkfs.fat:
- Use nicer error messages when encountering negative values in badblocks file
- Prevent signed integer overflow with excessively huge badblocks file
- Correctly count amount of bad blocks specified in badblocks file

Feel free to cherry pick commits which look good to you. You can find two proof of concepts in the commit messages to see the differences.

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -280,8 +280,9 @@ static int fill_mbr_partition = -1;	/* Whether to fill MBR partition table or no
 
 /* Function prototype definitions */
 
-static void mark_FAT_cluster(int cluster, unsigned int value);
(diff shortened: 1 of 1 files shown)


pali:
This remind me that I have some older change which improve and fix error message from mkfs.fat. Now I have pushed it into master branch.

Your changes looks good, the first is probably is obsolete now. Could you look at changes in master branch?

ferivoz:
> Your changes looks good, the first is probably is obsolete now. Could you look at changes in master branch?

I have removed the first commit and adjusted the second one.

PR #195, opened 2023-04-03
fsck.fat: Reduce duplicate file name lookups

ferivoz:
If a file has to be generated, then alloc_rootdir_entry tries to create a unique name by using a counter curr_num for name generation and looking for duplicates. If no duplicate is found, the counter curr_num is not incremented, which means that the next file uses the same name again. It is only incremented after the previously created file is detected, just to repeat the loop again.

The improvement is simple: Increment the counter after creating the filename. This has also the benefit that the check and use of curr_num are closer to each other within the source code.

### DIFF
--- a/src/boot.c
+++ b/src/boot.c
@@ -108,7 +108,10 @@ off_t alloc_rootdir_entry(DOS_FS * fs, DIR_ENT * de, const char *pattern, int ge
 	if (gen_name) {
 	    while (1) {
 		char expanded[12];
+		if (curr_num >= 10000)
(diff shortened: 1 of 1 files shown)


pali:
Looks good. I would suggest to do curr_num increment outside of the sprintf call, so sprintf does not have side-effect.

ferivoz:
> Looks good. I would suggest to do curr_num increment outside of the sprintf call, so sprintf does not have side-effect.

Thank you. I have added your suggestion to my commit.

PR #197, opened 2023-04-19
mkfs.fat: Fix signal race

ferivoz:
The signal handler used during bad block detection calls the signal-unsafe function printf. This could get in conflict with a printf for an error condition in do_check.

Just set a sig_atomic_t variable to indicate that current status should be displayed and leave the signal handler as soon as possible.

The main loop checks the flag and prints the status when needed.

### DIFF
--- a/src/mkfs.fat.c
+++ b/src/mkfs.fat.c
@@ -250,7 +250,6 @@ static int size_fat_by_user = 0;	/* 1 if FAT size user selected */
 static int dev = -1;		/* FS block device file handle */
 static off_t part_sector = 0; /* partition offset in sector */
 static int ignore_safety_checks = 0;	/* Ignore safety checks */
-static off_t currently_testing = 0;	/* Block currently being tested (if autodetect bad blocks) */
 static struct msdos_boot_sector bs;	/* Boot sector data */
 static int start_data_sector;	/* Sector number for the start of the data area */
 static int start_data_block;	/* Block number for the start of the data area */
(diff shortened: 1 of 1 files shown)


pali:
Hello! Thank you for a patch. I have a few comments:
* Global variable `currently_testing` is used only in check_blocks() function. So make this variable local.
* Comment about _Alarm clock handler_ is not accurate anymore.
* Explicitly initialize display_status to 0 before entering while-loop. I know that it is not needed (because of global variable) but it makes sense to do it because it is scoped variable just for check_blocks function (with its callback).
* Conversion from signal() to sigaction() is not related to _signal race_ change, so move it to separate commit (can be still part in this pull request) and ideally also unregister SIGALRM handler after while-loop.

pali:
And I found another signal/alarm race in the code. read() syscall in do_check() called from check_blocks() when alarm is installed can be interrupted. It either returns -1 (with errno EINTR) or fewer read data and in both case has to be called again.

Maybe it would make sense to protect all interruptable syscalls by recalling them again? This would allow to use debugger or stopping mkfs process...

ferivoz:
Hello @pali and thank you for your feedback!

I have split the changes into two commits. The first fixes the signal race, the other the portability issue with signal and also reducing the handler to a single variable assignment.

Are you sure that we have an issue with SIGALRM here? The SA_RESTART flag should avoid -1 return values and short reads.

pali:
Looks good. About signal safety, it looks like that SIGALRM with SA_RESTART should be safe there (we are not using network sockets, nor any other blocking ioctls). I'm not sure about other signals or usage of ptrace. But this is not related to the current PR.

pali:
Hm... I see there one issue, when sigaction() fails. Then calling of alarm() function is not a good idea.

ferivoz:
I have adjusted the line. The error is rather theoretical anyway, but I agree: Either handle it or do not handle it. Not in between.

pali:
I agree, thanks, merged!

PR #198, opened 2023-04-19
mkfs.fat: Better handling of bad blocks

ferivoz:
FAT32 filesystems may end up invalid when bad blocks are encountered.
This happens because the amount of free clusters is not adjusted.
Also if the root directory cluster is bad, then the filesystem cannot even
be repaired by fsck.fat.

These commits adjust mkfs.fat to find another valid cluster if the default
root cluster is invalid and updates tables accordingly.

Proof of Concept:

1. Create a FAT32 filesystem with bad root cluster:
```
echo 1592 > badblocks
mkfs.fat -C -F 32 -l badblocks poc.iso 102400
```

2. Check filesystem

```
fsck.fat poc.iso
```


pali:
Thank you for this change! Could you please write a test for this issue?

Also I see there an improvement for future changes - root directory for is written at fixed cluster - immediately after FAT tables. But for FAT32 it can be on any cluster as it is specified by _root_cluster_ member in boot sector. So _root_cluster_ should be chosen to the first non-bad cluster after FAT table. If it is not possible than mkfs.fat should throw an error. Also via _reserved sectors_ can be FAT table moved (in case there is a bad sector at the beginning of the disk).

And `struct fat32_fsinfo` definition could be improved in future too... Code is correct but slightly hard to read as structure does not match specification and so code is adjusted... The correct structure definition is `struct info_sector`.

ferivoz:
Thank you for the feedback @pali!

I have added a unit test and removed the Proof of Concept from the last commit message. It is not needed anymore, after all we have a unit test now.

The test itself also revealed what you have written: The label is always added to cluster 2 for FAT32, which of course is not correct if cluster 2 is marked with FAT_BAD. I have adjusted write_tables to reflect this as well.

Let me know what you think and if there is something else that could be simplified. As far as I understand your last remark, it should be done in another PR. Is this correct?

pali:
What I mentioned in the previous post is for another pull request.

For this pull request I would suggest to move code which prints "%d bad block%s\n" into some common function, maybe merge it together with your "update_tables" and call it from those check_blocks() and get_list_blocks(). And 
when it is called only from these bad blocks context, maybe renaming "update_tables" to "update_bad_blocks_info" (or something similar)?

And what is still missing there. If some bad block is _before_ FAT32 root dir or _before_ FAT12/16 data region then mkfs.fat must throw fatal error. As we cannot handle bad block in FAT table or in FAT16/16 root region. This should be simple check in that "update_bad_blocks_info" function (or how to call it).

I was thinking if it does not make sense to change order of setup_tables() and bad block processing functions. But when I'm looking at it again, seems that it is not needed.

And for another feature for future PR, adding support for running /sbin/badblocks directly from mkfs.fat (via popen()?); and adding support for reading bad blocks from existing FS when reformatting from existing FAT. I'm writing it here, so I do not forget about it :-) Nothing for this PR. Maybe it could be interested also for you.

ferivoz:
I have moved the function call and renamed it.

Please note that the checks are already in place. You can write a 1 into badblocks file and try to create a filesystem. If the bad blocks are outside the data area, a die message is already shown.

src/mkfs.fat.c
```diff
@@ -1372,6 +1393,35 @@ static void setup_tables(void)
...
+	info = (struct fat32_fsinfo *)(info_sector_buffer + 0x1e0);
+
+	/* reduce amount of free clusters accordingly */
+	free_clusters = le32toh(info->free_clusters);
+	if (free_clusters < bad_clusters)
+		die("too many bad clusters");
+	info->free_clusters = htole32(free_clusters - bad_clusters);
+
+	/* verify that root directory is intact */
+	root_cluster = 2;
```
pali, line 1415: I would suggest to read this value (as it is already filled) instead harcoding constant 2 on another place.
```suggestion
	root_cluster = le32toh(bs.fat32.root_cluster);
```

pali (commented):
> Please note that the checks are already in place.

I see... Anyway, one suggestion below...

src/mkfs.fat.c
ferivoz: Oh right, that makes sense. Applied and thank you for the reviews!

src/mkfs.fat.c
```diff
@@ -1422,8 +1472,14 @@ static void write_tables(void)
...
-    /* Write the root directory directly after the last FAT. This is the root
-     * dir area on FAT12/16, and the first cluster on FAT32. */
+    /* Write the root directory. On FAT12/16 it is directly after the last
+     * FAT. On FAT32 seek to root cluster. */
+    if (size_fat == 32) {
+	unsigned int cluster = le32toh(bs.fat32.root_cluster);
+	off_t sector = (off_t)start_data_sector +
+	    (cluster - 2) * bs.cluster_size * (sector_size / HARD_SECTOR_SIZE);
+	seekto(sector * HARD_SECTOR_SIZE, "root sector");
+    }
```
pali, line 1482: I must admit that this code is hard to understand, specially because of `start_data_sector` variable which is in strange units. In PR #182 there is a proposal to cleanup of it and in code do not use strange units (fixed 512 or 1024).

So what about avoiding usage of `start_data_sector` and write it for example as this?

```suggestion
    if (size_fat == 32) {
	unsigned int root_cluster = le32toh(bs.fat32.root_cluster);
	off_t root_sector = reserved_sectors + nr_fats * fat_length + (root_cluster - 2) * bs.cluster_size;
	seekto(root_sector * sector_size, "root sector");
    }
```

The best would be if `start_data_sector` is in sector_size unit...

ferivoz: I have adjusted the commit and also added an off_t cast to avoid integer overflow on large filesystems.

pali:
Thanks, merged.

PR #200, opened 2023-04-26
Fix vasprintf implementation

ferivoz:
A va_copy call must be followed by a va_end call. Also make sure that no return value of vsnprintf could lead to an integer overflow.


pali:
> A va_copy call must be followed by a va_end call.

For sure, this is a valid bug fix.

But I do not think that `(size_t)length` is fixing any issue there. Explicit cast to size_t has the only difference when length is MAX_INT (2^31-1). But when vsnprintf returns MAX_INT then there is some another bug in the code as it really should not return such huge number. So I do not see a reason for this explicit casting.

pali:
If any code in dosfstools wants to allocate 2GB of memory as a result of the formatting string/output then there is really something wrong. This really should not happen.

ferivoz:
> If any code in dosfstools wants to allocate 2GB of memory as a result of the formatting string/output then there is really something wrong. This really should not happen.

Okay. I have removed the casts from the commit.

pali:
Thanks, merged.

PR #201, opened 2023-04-28
mkfs.fat: Reduce visibility of static buffer

ferivoz:
The blkbuf array in check_blocks is not needed. It is always given as an argument to do_check. Keep the buffer there for easier access and reduced visibility in code.


pali:
Looks good, merged.

PR #218, opened 2025-07-02
Fix autoreconf with gettext 0.25

alyssais:
This fixes the following error that appeared when running autoreconf after updating to gettext 0.25:

	configure.ac:76: error: possibly undefined macro: AM_ICONV
	      If this token and others are legitimate, please use m4_pattern_allow.
	      See the Autoconf documentation.
	autoreconf: error: /nix/store/dvpiwvz7an7icljfscdi76h11c03cma4-autoconf-2.72/bin/autoconf failed with exit status: 1

It also means that, with any version of gettext, autoreconf should now be able to figure out that config.rpath is required, so the custom autogen.sh script should no longer be necessary.

The version of gettext given in AM_GNU_GETTEXT_VERSION() is picked quite arbitrarily based on what's likely to be available in distros, since gettext itself is not actually used here, just some supporting stuff from it.


configure.ac
```diff
@@ -15,8 +15,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 AC_INIT([dosfstools], [4.2+git])
+AC_CONFIG_MACRO_DIRS([m4])
```
mcprat, line 18: I believe defining `AC_CONFIG_MACRO_DIRS` has no effect here...

I would also experiment with the other two lines,
just to see if it's actually one or the other or both which are actually necessary.

alyssais: I'm now on gettext 1.0, and it does not seem like any of these changes are necessary any more.

Issue #210, opened 2024-05-02
Loss of maintainer/owner access to repo


katexochen:
> I got an email from @pali saying that their Github account is blocked and that (github) support won't acknowledge the situation.
So unless anyone else with admin rights to https://github.com/dosfstools/dosfstools shows up, the next step would be to fork?

_Originally posted by @bjornfor in https://github.com/dosfstools/dosfstools/issues/179#issuecomment-2003832455_

katexochen:
@bjornfor @jan-kiszka let's move this discussion here so it is better visible to other users.

katexochen:
Maybe @andreasbombe or @daniel-baumann still have access to the repo/org?

andreasbombe:
@katexochen Well I guess it's a good thing that I never formally left this project and still have owner access. You know, not for immediate action on my part, just to give a notice that I'm still able to do things here. But I have to get up to speed on what is going on.

katexochen:
@andreasbombe any update on this? Were you able to connect with @pali to restore their access? Or are you planning to pickup maintenance yourself?

bjornfor:
@andreasbombe: Hi, this is a friendly ping about the [previous message](https://github.com/dosfstools/dosfstools/issues/210#issuecomment-2242361586).

katexochen:
@andreasbombe And another ping

Issue #175, opened 2022-02-10
The user cannot access a file directory, with the message of "Contains a  cluster (xxx). Assuming EOF"


DouZhaolei:
**【Background】:**
I have a board, reset the board and run the dosfsck command once. then, The user cannot access a file directory.
The operations of "cd  ***" and "rm  ***" was failed.

-  I checked the FSCK logs and found that a large number of file corruption check records were recorded.

![picture1](https://user-images.githubusercontent.com/40568744/153370084-bf1253bb-dfcc-455a-81a5-d4e24a66ddb9.png)

- **I re-do the dosfsck command,the directory will be deleted.**  and the system will be ok ,The fsck logs are as follows:

![picture2](https://user-images.githubusercontent.com/40568744/153373880-f760328a-ded2-46bc-b558-5e806837ba83.png)

**【question】：**
The error information is in the following code segment in check_file():

```
if (!curEntry.value || bad_cluster(fs, curr)) {
	    printf("%s\n  Contains a %s cluster (%lu). Assuming EOF.\n",
		   path_name(file), curEntry.value ? "bad" : "free", (unsigned long)curr);
	    if (prev)
		set_fat(fs, prev, -1);
	    else if (!file->offset)
		die("FAT32 root dir starts with a bad cluster!");
	    else
		MODIFY_START(file, 0, fs);
	    break;
}
```
In the else branch，why not modify the file name[0] as DELETED_FLAG ? or I shoud run dosfsck with the operation of "V"?

PR #224, opened 2026-04-09
fsck.fat: Do not leave invalid empty directories

mvrable:
When a VFAT file system is corrupted so a directory points to a free cluster, `fsck.vfat` will truncate it to a completely empty directory (not even containing `.` or `..` entries).  That is invalid, and so a subsequent `fsck` run will delete the directory.

But it is better to do this all within a single run, so that `fsck` doesn't exit and leave the file system in an invalid state.  In particular, this can cause errors in the Linux file system driver (`error, corrupted directory (invalid entries)`, see `fs/fat/inode.c` in the Linux kernel) and lead to the file system being remounted read-only. It ought to be the case that after `fsck` runs successfully the Linux kernel can mount and modify the file system without error.  After this change that should be possible.

Also add a test with a file system image with this particular type of corruption: an image containing a single directory (`DIR`), but with the first copy of the FAT modified to indicate the cluster with the contents of `DIR` is free.


mvrable:
I believe this might be the same issue as described at https://github.com/dosfstools/dosfstools/issues/175, though I discovered and fixed it independently and only later noticed that other reported issue.

I also know that this fix might not be merged right away (because of https://github.com/dosfstools/dosfstools/issues/210) but figure it's better to have it posted here so it can be picked up eventually, or other users can see and make use of it if they wish.

